PR #1, opened 2017-08-12
remove service method instead of event emitter

ojones:


src/components/views/Causes.js
```diff
@@ -15,9 +15,8 @@ class Causes extends Component {
 
   removeCause(id){
     console.log('click', id);
-    socket.emit("causes::remove", id, { cascade: true }, (error, cause) => {
```
ojones: this event does not execute remove on server (not sure why it works for create and find).

src/components/views/Causes.js
```diff
@@ -15,9 +15,8 @@ class Causes extends Component {
...
-      console.log('Remove a cause', cause);
-    })
+    const causes = feathersClient.service('/causes');
```
ojones: wasn't sure about the best react place to put this so left it here for now close to .remove function that uses it

src/components/views/Causes.js
```diff
@@ -15,9 +15,8 @@ class Causes extends Component {
...
+    causes.remove(id).then(cause => console.log('Remove a cause', cause));
```
ojones: This method removes from server and emits event.  Don't fully understand but found this in docs: "every service will automatically become an instance of a NodeJS EventEmitter." (found in https://docs.feathersjs.com/api/services#service-methods and https://docs.feathersjs.com/api/services#feathers-functionality)

PR #3, opened 2017-08-25
Minor ui fixes for header on mobile

ojones:


ojones:
This is a simple solution for now.  Would rather have the hamburger (aka mobile menu icon with 3 horizontal lines) transform into and "x" once the mobile menu is open.  But this will fix the invisible menu issue at least for now.

PR #124, opened 2017-10-27
Update web3 packages so compile works again.

caseyhelbling:
Fixes error:
Failed to compile.

./node_modules/web3/packages/web3/src/index.js 
Module not found: Can't resolve 'web3-eth' in '/.../dev/giveth/giveth-dapp/node_modules/web3/packages/web3/src'

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15599 @@
+{
+  "name": "giveth-dapp",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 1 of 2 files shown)


perissology:
We can't change the web3 reference. That version contains a fix for a bug in the parity client pub/sub, which we rely on.

I can't seem to reproduce this. I did `rm -rf node_modules/web3*` and then `npm install` and everything compiled fine.

What OS are you using? And what node & npm version?

PR #125, opened 2017-10-27
update favicon file

vincentlg:
just replacing react default favicon by https://giveth.io/favicon.ico


ojones:
Thank you. Next time please use develop branch. Guess we need to make that more obvious. Will work on that. But most importantly, Thank You!

PR #163, opened 2017-11-25
Very light UI updates create/edit campaign

caseyhelbling:
Starting to standardizing how we use buttons on the interface. Specifically, submission buttons are on the right and cancel/back buttons are on the left. 

Create Campaign:
2017-11-25 at 3 09 pm

Update Campaign:
2017-11-25 at 3 08 pm


vsimetka:
Thanks for that, can you also update the DAC and Milestone component? I think in future there will be a template for these forms, but for now the changes needs to be propagated there as well. Also please keep the <-back button at the top of the form.

caseyhelbling:
Sounds good. I'll get the other components updated too.

Issue #193, opened 2017-12-05
All text displayed by the dapp needs to be checked for Spelling and Grammar errors.


bowensanders:
I proposed two other issues (#189 and #190) and in going over those two files, realized that there were many errors in each.

I will document the other errors I have find here, as well as reference this issue in the PRs for the fixes.

bowensanders:
Updates to DelegateButton.jsx

Fix to line 54: "pick another" to "pick another."
Fix to line 83: "The donation has been delegated," to "This donation has been delegated,"
Fix to line 88: "The donation has been delegated," to "This donation has been delegated,"
Fix to line 170: "delegate this donation to" to "delegate this donation to:"
Fix to line 174: "delegate this donation to" to "delegate this donation to:"

bowensanders:
Update FormsyImageUploader.jsx

Fix to line 63: "a png or jpeg." to "a png or jpg file."
Fix to line 66: "an png or jpeg" to "a png or jpg file."

bowensanders:
As per issue #193
_____
Update JoinGivethCommunity.jsx

Fix to line 31: "Great to see" to "It's great to see" ("It&#8217;s great to see")
Fix to line 45: "Great to see" to "It's great to see" ("It&#8217;s great to see")
Fix to line 66: "Great to see" to "It's great to see" ("It&#8217;s great to see")
Fix to line 80: "Great to see" to "It's great to see" ("It&#8217;s great to see")
Fix to line 80: "campaign!" to "campaign,"
Fix to line 81: "However" to "however,"

bowensanders:
As per issue #193

_____
Update MainMenu.jsx

Fix to line 54: "Any changes you're making" to "Any changes you have made"

bowensanders:
As per issue #193
_____
Update ShowTypeDonations.jsx

Fix to line 79: "Be the first, donate now!" to "Be the first to donate now!"

bowensanders:
As per issue #193
______
Update UnlockWallet.jsx

Fix to line 79: "Note: for security reasons your" to "Note: For security reasons, your"

bowensanders:
See also issues #189 and #190, as they were also related to this issue.

PR #196, opened 2017-12-05
Update FormsyImageUploader.jsx

bowensanders:
As per issue #193
_____
Update FormsyImageUploader.jsx

Fix to line 63: "a png or jpeg." to "a png or jpg file."
Fix to line 66: "an png or jpeg" to "a png or jpg file."

### DIFF
--- a/src/components/FormsyImageUploader.jsx
+++ b/src/components/FormsyImageUploader.jsx
@@ -63,7 +63,7 @@ class FormsyImageUploader extends Component {
           help="A picture says more than a thousand words. Select a png or jpg file."
           validations="minLength: 1"
           validationErrors={{
-            minLength: 'Please select an png or jpg file.',
+            minLength: 'Please select a png or jpg file.',
           }}
           required={this.props.isRequired}
         />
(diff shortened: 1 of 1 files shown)


bowensanders:
So would you rather it's jpg or jpeg? It isn't that big of a problem...

The "a" vs. "an" is definitely correct. It isn't a preposition, it's an article. :)

see: https://www.grammar.com/a-vs-an-when-to-use/

"A Historic Topic - A vs. An

Writers sometimes confuse the use of the articles a and an. We were all taught that a precedes a word starting with a consonant and that an precedes a word starting with a vowel (a, e, i, o, u, and sometimes y).

Here’s the secret to making the rule work: The rule applies to the sound of the letter beginning the word, not just the letter itself. The way we say the word will determine whether or not we use a or an. If the word begins with a vowel sound, you must use an. If it begins with a consonant sound, you must use a."

Could you let me know what part of it you would like changed?

vsimetka:
I don’t care, pick one and stick to it :).

bowensanders:
Okay, so then the file is OK as is. Correct?

Issue #216, opened 2017-12-15
Donate popup `lessThan` error message is unclear for < 0.5 eth wallet balance


mattstabeler:
![image](https://user-images.githubusercontent.com/237220/34038954-e3d57a76-e185-11e7-9541-aeda4d02ddaf.png)

The message suggests that the amount entered is more than the wallet balance, but the error message is triggered when the amount it less than the wallet balance minus 0.5 ETH:

![image](https://user-images.githubusercontent.com/237220/34039109-4a7ae9a0-e186-11e7-9c1f-c79f7494b290.png)

Perhaps the greaterThan value should reflect the error message, and be set to `0.09`. and the wallet balance should be the upper limit.

PR #217, opened 2017-12-15
Adjust error message thresholds to display messages correctly 

mattstabeler:
for wallet balances less than 0.5 ETH resolves Giveth/giveth-dapp#216


src/components/DonateButton.jsx
```diff
@@ -263,12 +263,12 @@ class DonateButton extends Component {
                   onChange={(name, value) => this.setState({ mewAmount: value })}
                   placeholder="10"
                   validations={{
-                    lessThan: wallet.getBalance() - 0.5,
-                    greaterThan: 0.00000000009,
+                    lessThan: wallet.getBalance(),
+                    greaterThan: 0.09,
```
caseyhelbling: Would `0.091` be a valid amount? If so `Minimum value must be at least Ξ0.1` is not correct. Could we make this `greaterThanEqual: 0.10`?

src/components/DonateButton.jsx
```diff
@@ -263,12 +263,12 @@ class DonateButton extends Component {
                   onChange={(name, value) => this.setState({ mewAmount: value })}
                   placeholder="10"
                   validations={{
-                    lessThan: wallet.getBalance() - 0.5,
-                    greaterThan: 0.00000000009,
+                    lessThan: wallet.getBalance() + (gas * 1000000000),
```
perissology: okay, a couple of things.

1. do you mind re-naming `gas` to `gasPrice`. I know you didn't name gas, but this would make everything more clear.
2. `1000000000` wasn't clear to me what it was. But I think you are converting the gasprice from `gwei` to `wei` which is correct. To make this more clear can you either use use a variable like `gwei = 1000000000` or preferably use `Web3.utils` `utils.toWei(gas, 'gwei')`.
3. we also need to take into account the amount of gas used by the tx. In `componentDidMount` we call `estimateGas`. We can use that value ( with a default of 600k incase the estimate gas fails) and multiply it by the gasPrice in wei. 600k is the default b/c w/ our current contracts, ~496k is the most gas used. This gives us a bit of wiggle room.
4. I think the gas needs to be subtracted from the walletBalance, not added.
5. `wallet.getBalance()` returns eth, but you are adding wei, we need to convert to the same units for any adding/subtracting, then to eth for the validator.

src/components/DonateButton.jsx
```diff
@@ -263,12 +263,12 @@ class DonateButton extends Component {
...
+                    greaterEqualThan: 0.1,
```
perissology: I don't see a reason to have a minimum value of .1 eth. That was $80 the other day. If we want to have a minimum value, maybe it should be the gas used in the tx? @vojtechsimetka what do you think?

PR #268, opened 2018-03-31
changed table-container overflow in _dashboard.css

geleeroyale:
Managed to build and run feathers and the DApp locally, however had some wallet problems, so I did not test much further than checking that the problem is fixed. 

Testing in console showed that also on smaller sizes scroll stays intact because the table is scrollable itself.

### DIFF
--- a/src/styles/_dashboard.scss
+++ b/src/styles/_dashboard.scss
@@ -1,7 +1,7 @@
 .dashboard-table-view {
 
   .table-container {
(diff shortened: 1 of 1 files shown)


vojtechsimetka:

Built with commit c26e2b4aa79a5876a35d0079a4f00f1493eb5880

https://app.netlify.com/sites/giveth-dapp/deploys/5abfd0e8fd0efa608f507019

PR #296, opened 2018-04-11
Fix image aspect ratio issues - #273

### LINKED ISSUE #273: Unifying the image aspect ratio
**Edit: See the task description https://github.com/Giveth/giveth-dapp/issues/273#issuecomment-379605135x**

The picture, when uploaded is always, zoomed in so that the top and bottom are cropped off... we need a better zooming strategy... maybe so that nothing is ever cropped off?

twmilli:
Fixes #273 
* Image uploading component lets user crop to 4:3 ratio (excludes avatar)
* DAC, Campaign, Milestone cards show 4:3 aspect ratio with max-width: 350 px
* Milestones without image show the Giveth logo


vojtechsimetka:

Built with commit 54f5c2af76539699908b36b27d03d8cb4a903df3

https://app.netlify.com/sites/giveth-dapp/deploys/5acd7b37792f895d9878c62c

vojtechsimetka:
Merging now to avoid horrible conflicts. However, there are three visual issues to be tackled:

- The cards should be justified to left. Right now there is justification to centre which breaks if there is just a few entries.
- The spacing between cards should remain constant. They should be more responsive potentially allowing more than 350px width.
- The performance of the resizing is horrible. The image should be resized once the cropped area changes but only after the mouse button is released. Consider removing the preview image and only keeping the cropper.


Example of spacing issues. The cards should be justified to left.
screen shot 2018-04-11 at 12 42 12

satyavh:
So I checked this PR and it breaks in a couple of ways:

1. Resizing increases CPU a lot, I easily hit 100% when dragging / moving the resizer around. It makes the app very unresponsive. I feel this is not acceptable at this stage. I suggest trying another package or disabling the resizing for now.

2. The UX of resizing is..sorry to say, kind of horrible. It takes way too much space and the design doesn't match the aesthetics of the dapp. Please come up with a better concept or ask somebody else to help design one before we merge.

3. When itemising a milestone, for each item you can add an image. This loads the same huge resizing UX  in the modal, making it kind of impossible to use the modal (especially because scrolling the modal also zooms the resizer).

4. And...the preview in the modal is linked to the preview in the main form. The logic must be isolated here. I had to reload my browser tab because now the dapp got really stuck because of mem leaks and cpu increases.

5. Cards look ok, but they became bigger. Please move back to the old size.  Especially for the campaign, where we list all the milestones within a campaign, the milestone cards should be way smaller (at the least how they used to be).

twmilli:
Hey, @satyavh thanks for the feedback.

1. Which PR are you looking at here?  I believe this issue is fixed in #299 but I could be wrong.
2. Again are you looking at #299 after we removed the preview (which I think made it look worse)?  Either way, I will definitely tweak the design to conform.
3. This is partially fixed in #299 and I'm working on it more now.
4. Sorry about this, we are disabling resizing for the modal so I think that will fix this. (let me know if you disagree)
5. I was under the impression that we wanted the cards to be responsive.  I can easily set a fixed width if that is what we want.  Also, the styling has been improved in #299 

@vojtechsimetka it might be helpful if you could chime in with your thoughts here.

vojtechsimetka:
@twmilli yeah @satyavh will check the other issue and write his feedback there. This one is closed.

satyavh:
Moved commenst here: https://github.com/Giveth/giveth-dapp/pull/299

PR #299, opened 2018-04-11
Unifying image aspect ratio 273 (Fixup to #296)

twmilli:
More work on #273 (See #296 for more detail)
* Fixes styling issues with cards.  Masonry was creating a bug so I took it out and added some css-grid.
* Spacing between cards is held constant at 10px
* Resizing performance fixed by only cropping on "cropend" events (when mouse button released).
* Removed preview image for resizing because it was clutter and not particularly useful
* Added optional resize prop to FormsyImageUploader so that milestone items do not show the resizer
* Fixed minor styling issue on milestone items (I think)


satyavh:
So I checked this PR and it breaks in a couple of ways:

1. Resizing increases CPU a lot, I easily hit 100% when dragging / moving / zooming the resizer around. It makes the app very unresponsive. I feel this is not acceptable. I suggest trying another package or disabling the resizing for now.

2. I feel we need to match the design of resizing with the aesthetics of the dapp. Removing the preview fixed already some of it. Is there a way to style the resizer? At the least:
-  giver the corners radius and a border matching the form. 
- Make the background white, because it is not transparently shown in the dapp

3. When itemising a milestone, for each item you can add an image. Before these fixed, this loaded the same resizing UX in the modal, making it kind of impossible to use the modal (especially because scrolling the modal also zooms the resizer).  

4. But now...it loads the new image not in the modal, but in the main form. The logic must be isolated here.

Cards look ok, but they became bigger. Please move back to the old size. Especially for the campaign, where we list all the milestones within a campaign, the milestone cards should be way smaller (at the least how they used to be).

vojtechsimetka:
One more request, can we move to 16:9 ratio instead of 4:3? Sorry for that.

twmilli:
Updates 👍 
* now using 16:9 ratio
* Width of cards slightly decreased
* Cropping for avatar
* Styling for cropper (Not sure exactly what we want here)
* Preview for milestone appears and maxes at 100x100

Notes after update:
* Cropping for avatar does not make it obvious it will appear as a circle.  May consider transitioning to cropper that handles this more natively.
* Found existing bug where you need to reload your browser after you change your avatar.

PR #307, opened 2018-04-16
Upgrade to React 16

ascariandrea:
I wanted a feedback about `react-input-token` cause it's the only component left using `react@15`, as reported by running `yarn list react`

```
$yarn list react
yarn list v1.6.0
warning Filtering by arguments is deprecated. Please use the pattern option instead.
├─ react-input-token@1.1.2
│  └─ react@15.6.2
├─ react-js-pagination@3.0.1
│  └─ react@16.2.0
└─ react@16.3.1
✨  Done in 1.27s.
```

The component seems to be unmaintained, so it could be rewritten and included in the code base as it is quite small.

Let me know what do you think

### DIFF
--- a/src/components/views/EditMilestone.jsx
+++ b/src/components/views/EditMilestone.jsx
@@ -35,6 +35,8 @@ import User from '../../models/User';
 import GivethWallet from '../../lib/blockchain/GivethWallet';
 import MilestoneItem from '../../components/MilestoneItem';
 import AddMilestoneItem from '../../components/AddMilestoneItem';
+import AddMilestoneItemModal from '../../components/AddMilestoneItemModal';
+
 
 BigNumber.config({ DECIMAL_PLACES: 18 });
 
@@ -50,8 +52,8 @@ BigNumber.config({ DECIMAL_PLACES: 18 });
  *    id (string): an id of a milestone object
  */
 class EditMilestone extends Component {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
 
     this.state = {
       isLoading: true,
@@ -231,7 +233,7 @@ class EditMilestone extends Component {
     });
   }
 
-  getEthConversion(date) {
+  getEthConversion = (date) => {
     const dtUTC = getStartOfDayUTC(date); // Should not be necessary as the datepicker should provide UTC, but just to be sure
     const timestamp = Math.round(dtUTC.toDate()) / 1000;
 
@@ -291,7 +293,7 @@ class EditMilestone extends Component {
     return 'Update Milestone';
   }
 
-  addItem(item) {
+  addItem = (item) => {
     this.setState({ items: this.state.items.concat(item) });
   }
 
@@ -312,7 +314,8 @@ class EditMilestone extends Component {
     };
   }
 
-  toggleItemize() {
+  toggleItemize = () => {
+    console.log(this.state.itemizeState);
     this.setState({ itemizeState: !this.state.itemizeState });
   }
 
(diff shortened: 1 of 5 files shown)


vojtechsimetka:

Built with commit 5a41bfeb0ae2d18b815748a6ee7d0b46f6ed2967

https://app.netlify.com/sites/giveth-dapp/deploys/5ad514c5dd6a544f73ce8f89

ascariandrea:
Thanks for the review @vojtechsimetka.
I'm not able to create the campaign locally cause the button is missing, can you point me in the right direction to make it appear?
Thanks

vojtechsimetka:
@ascariandrea sure. go to src/configuration and change the environment variable to develop. Then you can create a wallet and send some rinkeby eth to it. Alternatively you can use one of the keystore files from dapp/keystores folder. I suggest to use the first one.

ascariandrea:
@vojtechsimetka thank you, should be good now!

src/components/views/EditMilestone.jsx
```diff
@@ -229,7 +231,7 @@ class EditMilestone extends Component {
     });
   }
 
-  getEthConversion(date) {
+  getEthConversion = (date) => {
```
vojtechsimetka: This is strange to me, can you revert to normal class member functions?

src/components/views/EditMilestone.jsx
```diff
@@ -746,13 +749,13 @@ class EditMilestone extends Component {
...
                         defaultChecked={this.state.itemizeState}
-                        onChange={() => this.toggleItemize()}
+                        onChange={this.toggleItemize}
                         disabled={!isNew && !isProposed}
                       />
                       <span className="label">Add multiple expenses, invoices or items</span>
                     </div>
 
-                    {!itemizeState && (
+                    {!itemizeState ? (
```
vojtechsimetka: This is alright but unnecessary modification.

src/components/views/EditMilestone.jsx
ascariandrea: Why is it strange? It avoids to bind the class members in the constructors, using the lexical scope.

src/components/views/EditMilestone.jsx
ascariandrea: Agree, but it prevents to render nullable element.

src/components/views/EditMilestone.jsx
vojtechsimetka: Sure that is clear, but linters don't like it and in general it seems to be discouraged (react, airbnb libs are not accepting code like that it seems)

src/components/views/EditMilestone.jsx
vojtechsimetka: 👍

src/components/views/EditMilestone.jsx
ascariandrea: As reported in the [documentation](https://reactjs.org/docs/handling-events.html), this is not discouraged, but I would say is the *old* way oof doing it, as the *fat arrow* is an *ES6* feature.

Anyway, I'll revert the changes cause the linter still complains and maybe we want to discuss it further.

ascariandrea: I don't consider that an issue: how many times you need to subclass a component? I think in React hemisphere is more about composition than subclassing.
The benchmarks show the ops/s for instantiating 100 times the same component, which in React is not that common, and when it happens the instantiate component is usually stateless, which it means that can be written as a function and have no method binding at all.

ascariandrea:
🎉

PR #319, opened 2018-04-23
Prettified and lint fixing on commit 

ascariandrea:
After few feedbacks received on this [thread](https://www.loomio.org/d/rEWpfYNg/pretty-and-lint-fix-on-commit), here I'm proposing the implementation.

For what I've noticed the lint fixing step could be faster, but it's just a matter of a bunch of seconds.

Feel free to give it a try and let me know your thoughts!


vojtechsimetka:

Built with commit 397ce3d9245d0150d33ae73f0344a3d9029af08a

https://app.netlify.com/sites/giveth-dapp/deploys/5addb18711b73b5a307fd2e1

vojtechsimetka:
Does it stop on warnings? If not then I suggest leaving the console statements as warnings without eslint disable. They are constant reminder those parts should change (and many have been/will be removed in  #310)

ascariandrea:
> does the linting only run on staged files? If possible, that may speed it up

Yep, the linting runs only on staged files (now).

@vojtechsimetka yep, it stops on warnings as well, even though I think `console.log` should be treated as errors from linting while the appropriate `TODO` comment should be used instead.

PR #323, opened 2018-04-25
Feature/user and wallet provider

rudin:
Here's an implementation where I moved the currentUser, wallet and the accompanied actions to a context Provider and Consumer pair.
The next step: Right now the UserConsumer is being used inside the 'Application', where the currentUser, wallet and actions are being passed down as props. I'd suggest using this consumer in the various screens and components where needed.


vojtechsimetka:

Built with commit 49ceeb895d3725ac0d989311169ca81631d6067e

https://app.netlify.com/sites/giveth-dapp/deploys/5ae07c27b13fb13c71a5c59a

Issue #329, opened 2018-04-26
Back button uses calls on browser back function


krrisis:
non-urgent issue but can be annoying during demos: 

When you link to a Milestone or a Campaign from somewhere (like here below) the 'back' arrow does not function, as it seems call on the back function of the browser. When you click the below link it opens in a new tab, so when you hit 'back' nothing happens, instead of going to the campaign to which this milestone is linked. 
eg: click here on the back arrow, nothing happens: https://alpha.giveth.io/campaigns/R1WkS0obautijkvJ/milestones/Oo3yG9msdYA78X78

RnbWd:
I can work on this

RnbWd:
I have a fix to this issue with the back button that goes to a lower route based on the URL. This can be implemented as the default behavior for the back button, but when adding new Routes, the functionality of the back button will have to be considered to prevent users from being stuck at a dead urls. The only dead url I could find was: `/campaigns/:id/milestones`, so I added a redirect Route for it to go to `/campaigns/:id/milestones`

vojtechsimetka:
This was fixed in the #403

Issue #182, opened 2017-12-02
Accessing protected URLs redirects user back to page outside of Giveth


vojtechsimetka:
If you directly access protected link while already being on the page, it will just redirect you back.

**Replicate:**
- Go to website like swarm.city
- Paste in the browser prodected dapp URL e.g. https://alpha.giveth.io/delegations
- Enjoy the loop

**Correct Behaviour:**
If the history.back points outside of giveth platform, it should redirect to the base url instead.

status-open-bounty:
![Contract deploying](https://openbounty.status.im/img/deploying_contract.png)
Transaction: [0x8c8780c9e9e4c8da3543577aa8814b51de7bf90390fa64122a23abb232c73c67](https://etherscan.io/tx/0x8c8780c9e9e4c8da3543577aa8814b51de7bf90390fa64122a23abb232c73c67)

RnbWd:
I have a fix for this issue

rsercano:
It's merged but still seems open in open bounty.

RnbWd:
hmm?

vojtechsimetka:
@RnbWd can you try an make empty PR with link to the original one and mention "fix #182" in it? I'll merge and that should do the trick.

vojtechsimetka:
@geleeroyale Can we pay this out somehow so that I can close?

geleeroyale:
wow - hurray for github notifications - sorry - I just did not see this. @RnbWd - please follow these steps:

1 - register on openbounty.status.im to link your github account to a metamask account
2 - sumbit an empty PR with the following comment: fixes #[your issue number] (without the square brackets)
3 - Ping @geleeroyale on Riot, to accept the PR

I will then close this issue again

PR #337, opened 2018-04-27
redirect to homepage if not authenticated, added redirect route for /…

RnbWd:
…campaigns/:id/milestones to prevent dead url, modify back button behavior

fixes https://github.com/Giveth/giveth-dapp/issues/329, https://github.com/Giveth/giveth-dapp/issues/182,

### DIFF
--- a/src/lib/middleware.jsx
+++ b/src/lib/middleware.jsx
@@ -10,8 +10,8 @@ import { history } from '../lib/helpers';
  * @return new Promise
  *
  * usage:
- *    isAuthenticated(currentUser, wallet)
- *      .then(()=> ...do something when authenticated)
+ *    isLoggedIn(currentUser)
+ *      .then(()=> ...do something when logged in)
  *      .catch((err) ...do something when not logged in
  *      returns new Error 'notLoggedIn' if not logged in
  */
--- a/src/components/views/EditCampaign.jsx
+++ b/src/components/views/EditCampaign.jsx
@@ -98,6 +98,12 @@ class EditCampaign extends Component {
         } else {
           this.setState({ isLoading: false });
         }
+      })
+      .catch(err => {
+        if (err === 'notAuthenticated') {
+          // if not authenticated
+          // history.push('/');
+        }
       });
   }
 
--- a/src/components/views/MyCampaigns.jsx
+++ b/src/components/views/MyCampaigns.jsx
@@ -36,7 +36,8 @@ class MyCampaigns extends Component {
       })
       .catch(err => {
         if (err === 'notLoggedIn') {
-          // history.push('/');
+          // not logged in
+          // notLoggedInPopup();
         }
       });
   }
--- a/src/components/views/MyDACs.jsx
+++ b/src/components/views/MyDACs.jsx
@@ -28,13 +28,19 @@ class MyDACs extends Component {
   }
 
   componentDidMount() {
-    isLoggedIn(this.props.currentUser).then(() => {
-      this.dacsObserver = DACservice.getUserDACs(
-        this.props.currentUser.address,
-        dacs => this.setState({ dacs, isLoading: false }),
-        () => this.setState({ isLoading: false }),
-      );
-    });
+    isLoggedIn(this.props.currentUser)
+      .then(() => {
+        this.dacsObserver = DACservice.getUserDACs(
+          this.props.currentUser.address,
+          dacs => this.setState({ dacs, isLoading: false }),
+          () => this.setState({ isLoading: false }),
+        );
+      })
+      .catch(err => {
+        if (err === 'notLoggedIn') {
+          // not logged in
+        }
+      });
   }
 
   componentWillUnmount() {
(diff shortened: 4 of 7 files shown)


vojtechsimetka:

Built with commit 5da8561210d6320169da585394ec92f3c2b98652

https://app.netlify.com/sites/giveth-dapp/deploys/5ae37ed111b73b0dadd0ec28

vojtechsimetka:

Built with commit 5da8561210d6320169da585394ec92f3c2b98652

https://app.netlify.com/sites/giveth-dapp/deploys/5ae37ed23672df050f61bf1e

src/components/GoBackButton.jsx
```diff
@@ -3,8 +3,32 @@ import PropTypes from 'prop-types';
 
 const GoBackButton = ({ history, styleName }) => (
   <div
-    onClick={history.goBack}
-    onKeyPress={history.goBack}
+    onClick={() => {
+      // ths back button will go a lower nested route inside of the DApp
+      // if the browser history is zero
+      //  this should be the default behavior of the back button
+      if (history.length < 1) {
```
vojtechsimetka: Could you extract the logic that is repeated in both `onClick` and `onKeyPress` handlers to a function? Also I don't think this solves the issue of accessing giveth from external site. I think the back button should always go one route lower and if the user wants to go back they use the browser back button.

src/lib/middleware.jsx
```diff
@@ -16,7 +16,7 @@ import { history } from '../lib/helpers';
 export const isLoggedIn = currentUser =>
   new Promise(resolve => {
     if (currentUser && currentUser.address) resolve();
-    else history.goBack();
+    else history.push('/');
```
vojtechsimetka: This is super ugly to have it in the isLoggedIn user. Could this just reject the promise and then the functionality be handled outside? I may even make sense to inform the user that they were accessing protected URL and should login first. What do you think?

src/lib/middleware.jsx
```diff
@@ -35,7 +35,7 @@ export const isLoggedIn = currentUser =>
 export const isAuthenticated = (currentUser, wallet) =>
   new Promise(resolve => {
     if (currentUser && currentUser.address && wallet && wallet.unlocked) resolve();
-    else history.goBack();
+    else history.push('/');
```
vojtechsimetka: Same feedback as in isLoggeIn function.

src/components/GoBackButton.jsx
RnbWd: yes i agree, I can extract the logic into a utility function, without the conditional statement which isn't necessary

src/lib/middleware.jsx
RnbWd: Yes, I agree that makes a lot of sense, but in the meantime this is a better solution

RnbWd:
> @vojtechsimetka I've noticed one more thing, there is a back button when creating/proposing/editing milestone, however that one does not have a campaign in the URL. Could you change the URLs as well?

I noticed that the milestone path in the main application route was structured like that, but it also looked like you could edit and create milestones from a url with a campaign, and that's what I saw when testing, are both paths for milestones identical?

vojtechsimetka:
Yes please do

vojtechsimetka:
@RnbWd are you still working on this? Is there something we can help you with?

RnbWd:
Yes, the changes are pretty simple, but I was wondering what we'd use to notify the user they were accessing a protected URL, a swal popup?

vojtechsimetka:
Hi @RnbWd, is this something you are still actively pursuing?

RnbWd:
I added a popup that either redirects home or to the sign in page if the user is not logged it, and it rejects the promise. I haven't gone through the codebase yet to make sure the promise rejections are handled properly. This PR also changes the functionality of the back button (to go back one lower nest in the url instead of using the browser's default goback behavior)

RnbWd:
the promises are handled okay.. I the popup dialogue doesn't open in the catch statement, so the promise rejection doesn't do much, but its there in case anyone wants to use that function

RnbWd:
made a new pull request https://github.com/Giveth/giveth-dapp/pull/403

satyavh:
Merged this, some feedback:

I reviewed this PR but  the solution only works if the user is actually signed out. If the user is signed in, and you do a browser refresh or go to a restricted page directly, this solution fails.

1. The issue is that when you are signed in, and go to for example `/my-dacs`, the popup shows, but at the same time in the background it loads the DACs.  I'm signed in, so why does the popup show at all? It should not.

The problem is not this PR though (although it should have been tackled). There's a widespread issue within the dApp that re-renders and remounts components. For example, for the MyDACs component, `componentDidMount` is triggered 4 (!) times. That results in this code not working properly:

`isLoggedIn(this.props.currentUser)` 

Because the currentUser is undefined for the first, second and third mount, but the 4th mount the user is defined. However the first mount React.swal is already triggered, and then for the 4th mount the data is loaded in the background.

I've made a fix for that in the `UserProvider` and now the popup works correctly. Yeah! :-)

2. When you are on a restricted route, and press the sign out button, you get to see the pop-up. That's unnecessary information. So I've made a fix to route away from the page.

Apart from these 2 issues, great job! The new implementation of the GoBackButton is helping a lot with other issues we're having. So thanks a lot for tackling this!!

Issue #336, opened 2018-04-26
Issue with fiat amount in this Milestone


krrisis:
Shows up as 'undefined undefined'

https://alpha.giveth.io/campaigns/OcKJryNwjeidMXi9/milestones/iMrCZbvlmFvjAgKA

ascariandrea:
I can work on this!

vojtechsimetka:
There are values missing in the items array for this milestone:

- selectedFiatType
- fiatAmount
- conversionRate
- etherAmount

ascariandrea:
The object creation should be prevented when no such data are available on item creation

vojtechsimetka:
Should be handled in ff11e4508222de587c47cff2ed94fc30b4ed2ce6

PR #340, opened 2018-05-01
Added missing keys to render MilestoneItem

ascariandrea:
#336

Maybe I'm wrong, but I think there's a missing `etherAmount` key in the response received from the API.

Can someone please confirm this?

Or tell me I'm stupid is also an option :)

### DIFF
--- a/src/components/views/ViewMilestone.jsx
+++ b/src/components/views/ViewMilestone.jsx
@@ -166,6 +166,7 @@ class ViewMilestone extends Component {
       status,
       fiatAmount,
       selectedFiatType,
+      conversionRate,
       campaign,
       campaignOwnerAddress,
     } = this.state;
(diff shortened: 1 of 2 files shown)


vojtechsimetka:

Built with commit 8f380962a6cc4a16bc66eea3d0ec5ff92ce82eb6

https://app.netlify.com/sites/giveth-dapp/deploys/5ae8a0428df8947532c12dda

ascariandrea:
I reverted my changes and also drop the `develop` branch merging commit.

But my question is still unanswered :/

Should the `item` have `etherAmount` key?

vojtechsimetka:
I don't have time to properly check now, but I think so. Look at the Rate Convertor component (which should be actually used in the milestone edit itself because we are just repeating the logic).

src/components/views/ViewMilestone.jsx
```diff
@@ -255,7 +254,19 @@ class ViewMilestone extends Component {
...
-                                <MilestoneItem name={`milestoneItem-${i}`} item={item} />
+                                <MilestoneItem
+                                  key={item.date}
+                                  name={`milestoneItem-${i}`}
+                                  item={{
+                                    date: item.date,
+                                    description: item.description,
+                                    selectedFiatType,
+                                    fiatAmount,
+                                    conversionRate,
```
vojtechsimetka: The selectedFiatType, fiatAmount and conversionRate are surely (and if not should be) defined and stored per item.

vojtechsimetka:
So the way milestones work now is:

- Milestones can from now on be denominated in one token or ETH.
- Each item can however have a conversion from any fiat currency or ETH.

ascariandrea:
I fixed using the item's values for `MilestoneItem` rendering

PR #356, opened 2018-05-22
[FIX #297] Nodejs: possible memory leak detected

### LINKED ISSUE #297: Nodejs: possible memory leak detected
When using the DApp, the following warning is logged:

```
events.js:138 (node) warning: possible EventEmitter memory leak detected. 31 listeners added. Use emitter.setMaxListeners() to increase limit.
a.addListener @ events.js:138
```
When compiling the DApp, the following warning is logged:

```
(node:8744) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGINT listeners added. Use emitter.setMaxListeners() to increase limit
```

ovx:
Fixes #297 

Solution: implement a cache mechanism for `MiniMeToken` instances in the `getBalance` function which is executed in intervals. Each `MiniMeToken` instance is identified by its wallet address.

Problem description: The memory leak warning is related to the web3's `web3-eth-contract` submodule which executes `setProvider` on every contract creation and binds `data` event on the global instance of `Web3ProviderEngine`. The method `getBalances` https://github.com/Giveth/giveth-dapp/blob/develop/src/lib/blockchain/getWeb3.js#L42 is executed in intervals (15s) and creates a new instance of the contract on each execution and binds a new `data` event on the global provider. The caching mechanism re-uses already created contracts and thus doesn't bind unnecessary events. 

Reproduction: Change `setInterval` (https://github.com/Giveth/giveth-dapp/blob/develop/src/lib/blockchain/getWeb3.js#L42) to about 500ms, soon you will see the eventemitter warning.

I could not reproduce the other warning `SIGINT` when compiling. Because this other warning is certainly related to one of the build processors or react tools, updating  these tools to newer versions should fix it it it's still present.

### DIFF
--- a/src/lib/blockchain/getWeb3.js
+++ b/src/lib/blockchain/getWeb3.js
@@ -39,19 +39,25 @@ function setWallet(wallet) {
     },
   });
 
+  const minimeTokenCache = {};
+
   const getBalance = () =>
     getWeb3() // eslint-disable-line no-use-before-define
       .then(web3 => {
         const { tokenAddresses } = config;
         const addr = wallet.getAddresses()[0];
 
-        const tokenBal = tAddr =>
-          tAddr
-            ? new MiniMeToken(web3, tAddr).balanceOf(addr).then(bal => ({
-                address: tAddr,
-                bal,
-              }))
-            : undefined;
+        const tokenBal = tAddr => {
+          if (tAddr) {
+            const token = minimeTokenCache[tAddr] || new MiniMeToken(web3, tAddr);
+            minimeTokenCache[tAddr] = token;
+            return token.balanceOf(addr).then(bal => ({
+              address: tAddr,
+              bal,
+            }));
+          }
+          return undefined;
+        };
         const bal = () => (addr ? web3.eth.getBalance(addr) : undefined);
 
         return Promise.all([bal(), ...Object.values(tokenAddresses).map(a => tokenBal(a))]);


vojtechsimetka:

Built with commit ef428aafa7dc27f746e936fc7f0d6e5517834f13

https://app.netlify.com/sites/giveth-dapp/deploys/5b04208bdd6a544c19c7b28d

vojtechsimetka:

Built with commit ef428aafa7dc27f746e936fc7f0d6e5517834f13

https://app.netlify.com/sites/giveth-dapp/deploys/5b04208bb312744e5526f5f6

perissology:
Is it possible to fix this in web3? It seems like we should be able to create contract instances w/o attaching too many listeners.

This fix won't work if someone does 30 (or whatever the maxListenersLimit is) interactions with dac/campaigns/milestones/donations before refreshing the page, unless we cache every contract they interact with. This would be a bit annoying.

ovx:
@perissology There's nothing really wrong with web3, the bug is clearly in giveth

It is important to understand that this warning is just a warning and it exists for a good reason. Its purpose is to warn developers that they are binding too many events and this could indicate a memory leak. It doesn't necessarily mean that is a memory leak, all this warning is saying is just that you should be more careful.

In this specific case, I think you are in fact create a memory leak. With each new contract instance, you are creating a new instance of `RequestManager` and looking at the implementation here: https://github.com/ethereum/web3.js/blob/v1.0.0-beta.34/packages/web3-core-requestmanager/src/index.js#L98 this cannot be even properly garbage collected because it keeps references to the global provider and these instances of RequestManager will stay in memory. This is something web3 could do better, but this is simply the way web3 works and its indended use. Technically nothing wrong with it if you know how to use it.

Imho you should cache every contract users interact with, this is the intended use of web3's API - contracts are just interfaces and there is no reason to create a new instance for every use. (Anyone is welcome to correct me here, if my understanding of web3 API is wrong). 

Either way, I think web3's implementation is OK and I'm not going to push some obscure workarounds into web3 just because you're using it wrong.

If you think warnings about potential issues and wrongdoings are annoying, you can always disable this warning by adding `engine.setMaxListeners(0)` to https://github.com/Giveth/giveth-dapp/blob/develop/src/lib/blockchain/getWeb3.js#L80. This way you can create as many memory leaks as you want ;)

perissology:
I want to thank you again for looking into this. However after discussing this as a team, we believe that this is really a bug with web3.

Your fix does address a hack in the giveth code, and I've included the commit in the develop branch. However it doesn't fix the problem at the root. We don't think a fix in web3 would be an obscure workaround, but instead it would be a proper implementation.

For you're contribution, we are awarding you 200 DApp points [more info](https://medium.com/giveth/how-rewarddao-works-aka-what-are-points-7388f70269a)

ovx:
@perissology I understand but I still stand by my opinion that the problem is how you use web3, not in web3 itself. You'll have the same issues in many other places and with other events unless you introduce some big changes into web3. 

And thank you for the DApp points. How to claim them? Whom should I contact?

vojtechsimetka:
Hi @ovx , would you mind joining our https://riot.im room? It would make everything easier  #giveth-product-development:matrix.org . I will dish those points to you and we would like to ask you to make a video about your contribution with your ETH address as described here https://medium.com/giveth/how-rewarddao-works-aka-what-are-points-7388f70269a . You will then receive your reward in ETH around 20th of June.

ovx:
@vojtechsimetka thanks for the info, I joined the riot group

PR #359, opened 2018-05-25
feature/milestone-conversation

satyavh:


vojtechsimetka:

Built with commit 8526ff04bc59de294cf6218695fcbc697de3666a

https://app.netlify.com/sites/giveth-dapp/deploys/5b0845071f12b7212382d1f8

vojtechsimetka:

Built with commit 8526ff04bc59de294cf6218695fcbc697de3666a

https://app.netlify.com/sites/giveth-dapp/deploys/5b084507b3127475c26a52de

src/lib/helpers.jsx
```diff
@@ -106,6 +106,10 @@ export const getGasPrice = () =>
       // we're only interested in gwei.
       // we round to prevent errors relating to too many decimals
       gasPrice = Math.round(gasPrice) / 10;
+
+      // sometimes the API is down, we need to return a gasprice or the dapp breaks
+      if (!gasPrice) gasPrice = 10;
```
vojtechsimetka: Maybe feathers should always return the gas price and not DApp? If not could you extract this as constant?

src/components/MilestoneConversations.jsx
```diff
@@ -0,0 +1,110 @@
...
+import moment from 'moment';
+
+import { feathersClient } from './../lib/feathersClient';
+import Loader from './Loader';
+import { getUserName, getUserAvatar } from './../lib/helpers';
+import getNetwork from './../lib/blockchain/getNetwork';
+
+/* eslint no-underscore-dangle: 0 */
+class MilestoneConversations extends Component {
+  static getReadeableMessageContext(context) {
```
vojtechsimetka: Would make sense to name this `state` or `milestoneState` to avoid confusion.

vojtechsimetka: Also note to self: Replace with constants from the milestone model once rewriting.

satyavh: The reason I've called it context is because the context of a conversation message isn't always the milestone state. For example, there's currently a context called 'replyTo' (not implemented in frontend) and perhaps in the future there will be other contexts.

Agree about replacing with constants

src/lib/helpers.jsx
satyavh: The feathers API could be down as well :-)

So the dApp needs to work with a default gas price. I've made it a constant in configuration.

PR #360, opened 2018-05-25
UI "done" for very simple splitting. 

vojtechsimetka:
Need to fix feathers which is splits the donations incorrectly (the donation that holds the `originalAmount-delegatedAmount` also appears to be delegated but should not be)

### DIFF
--- a/src/models/Donation.js
+++ b/src/models/Donation.js
@@ -30,6 +30,18 @@ class Donation extends Model {
     return 'cancelled';
   }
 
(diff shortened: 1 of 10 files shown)


src/models/Donation.js
```diff
@@ -28,13 +30,25 @@ class Donation extends Model {
     return 'cancelled';
   }
 
+  static get statuse() {
```
perissology: typo

Issue #368, opened 2018-05-29
List of supported exchange rates should be retrieved from feathers service


vojtechsimetka:
Rather than having the list of supported exchange rates hardcoded in [`/src/components/RateConvertor.jsx`](https://github.com/Giveth/giveth-dapp/blob/develop/src/components/RateConvertor.jsx) and [`src/components/views/EditMilestone.jsx`](https://github.com/Giveth/giveth-dapp/blob/develop/src/components/views/EditMilestone.jsx) it should come from feathers.

wildaces215:
Is this open?

vojtechsimetka:
Yes

wildaces215:
Ok I would like to take it! Should the data come from an api?

wildaces215:
That way more currencies can be supported

vojtechsimetka:
For the time being I think this list needs to be exposed in feathers and read in UI.
https://github.com/Giveth/feathers-giveth/blob/abdf973189977ab57dbd0568c4410ca2d7a1fbed/src/services/ethconversion/getEthConversionService.js#L4

Satya will rewrite that part soon because it can not fetch historical prices in the current implementation.

wildaces215:
Ok sounds good!

perissology:
looks like this is done

PR #382, opened 2018-06-12
Addressed issue #368

wildaces215:
I created a feathers service for the various currencies, now I am curious what functions should be added to the service in order for it to be more efficient and for the getFiatType function should it return a promise?


vojtechsimetka:

Built with commit 066d9068be6bbd5ddaef0a41ceb437820df78d79

https://app.netlify.com/sites/giveth-dapp/deploys/5b1fd580b312744f0479a95d

vojtechsimetka:

Built with commit 066d9068be6bbd5ddaef0a41ceb437820df78d79

https://app.netlify.com/sites/giveth-dapp/deploys/5b1fd581b13fb1746c716f9a

package.json
```diff
@@ -59,9 +59,12 @@
     "eslint": "^4.10.0",
     "eslint-config-airbnb": "^16.1.0",
     "eslint-config-prettier": "^2.9.0",
+    "eslint-config-standard": "^11.0.0",
     "eslint-loader": "^1.9.0",
+    "eslint-plugin-babel": "^5.1.0",
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jsx-a11y": "^6.0.2",
+    "eslint-plugin-node": "^6.0.1",
```
vojtechsimetka: Do we need these?

src/services/Currency.js
```diff
@@ -0,0 +1,20 @@
+import { feathersClient } from '../lib/feathersClient';
+import { resolve } from 'dns';
+
+class Currency {
+  fiatTypes = ['BRL', 'CAD', 'CHF', 'CZK', 'EUR', 'GBP', 'MXN', 'THB', 'USD'];
```
vojtechsimetka: This needs to be array of objects { value, title } or you need to change how the fiat types are used.

src/services/Currency.js
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * Get fiat type
+   * @param id
+   */
+
+  getFiatType(id) {
```
vojtechsimetka: This function isn't used anywhere, I don't understand what it really does and why it is here.

wildaces215:
Hey can I make the changes? And I also ask for help I'm not very familiar with feathers js. Also the reason why I create the function getFiatType() was to retrieve the fiat type itself form a function unless you want to the import the array itself.

vojtechsimetka:
No need for a function, you can just import the whole array similar to how it is done here. https://github.com/Giveth/giveth-dapp/blob/ec3d377013a537f130efbdeea3d845a1f1d12259/src/contextProviders/UserProvider.jsx#L24

I'm sorry right now I can not give you guidance on the feathers part, we have other priorities and a big release coming up. However, I am working on fixing something else in that service so I may write the feathers part for you.

wildaces215:
Ok I understand will I'll do is do what you said about the array and add it to currency.js and do another pull request and let you do what you need todo.

vojtechsimetka:
Hi @wildaces215 there is a lot of other files that are modified that did not need to be modified. E.g. configuration.js, DAC.js etc. Can you make this the most minimal modification possible?

wildaces215:
I think I'll make a fresh fork of giveth for some reason I only wanted to
change one thing but you cancel this pull request. I'm sorry about the
inconvenience

PR #456, opened 2018-08-15
Feature/donation aggregation

### LINKED ISSUE #84: Allow delegating money from Donate button in Campaigns and Milestones
If you are a delegate and you have money to delegate, clicking on the Donate button (or GivETH) should offer you the option to delegate the DAC money.

vojtechsimetka:
Fixes #84

### DIFF
--- a/src/components/DelegateMultipleButton.jsx
+++ b/src/components/DelegateMultipleButton.jsx
@@ -99,7 +99,11 @@ class DelegateMultipleButton extends Component {
               ])
             : dacs;
 
-          this.setState({ delegationOptions });
+          this.setState({ delegationOptions }, () => {
+            if (delegationOptions.length === 1) {
+              this.selectedObject({ target: { value: [delegationOptions[0].id] } });
+            }
+          });
         },
         () => {},
       );
--- a/src/components/DelegateMultipleButton.jsx
+++ b/src/components/DelegateMultipleButton.jsx
@@ -157,9 +157,7 @@ class DelegateMultipleButton extends Component {
         r => {
           const delegations = r.data.map(d => new Donation(d));
           let amount = utils.fromWei(
-            delegations
-              .reduce((sum, d) => sum.plus(new BigNumber(d.amountRemaining)), new BigNumber('0'))
-              .toString(),
+            delegations.reduce((sum, d) => sum.add(utils.toBN(d.amountRemaining)), utils.toBN('0')),
           );
 
           if (
--- a/src/components/DelegateMultipleButton.jsx
+++ b/src/components/DelegateMultipleButton.jsx
@@ -1,0 +1,347 @@
+import React, { Component } from 'react';
+import Modal from 'react-modal';
+import { utils } from 'web3';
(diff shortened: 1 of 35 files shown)


vojtechsimetka:

Built with commit 439fcb8b2b81bd8baaec4615ea45bf861a00fe17

https://app.netlify.com/sites/giveth-dapp/deploys/5b7432963813f025edac0920

src/components/DelegateMultipleButton.jsx
```diff
@@ -0,0 +1,363 @@
...
+    this.donationsObserver = feathersClient
+      .service('donations')
+      .watch({ listStrategy: 'always' })
+      .find(query)
+      .subscribe(
+        r => {
+          const delegations = r.data.map(d => new Donation(d));
+          let amount = utils.fromWei(
+            delegations
+              .reduce((sum, d) => sum.plus(new BigNumber(d.amountRemaining)), new BigNumber('0'))
```
perissology: just an fyi, but if you use utils.toBN(..) instead of new BigNumber, you can pass the BN object to fromWei & don't have to call .toString()

src/contextProviders/DonationProvider.jsx
```diff
@@ -56,6 +56,7 @@ class DonationProvider extends Component {
             query: {
               giverAddress: this.props.currentUser.address,
               amountRemaining: { $ne: 0 },
+              pendingAmountRemaining: { $ne: 0 },
```
perissology: why was this added? Doesn't this do an AND query

src/models/Campaign.js
```diff
@@ -56,6 +57,10 @@ class Campaign extends BasicModel {
     return campaign;
   }
 
+  get type() {
+    return this._type;
+  }
+
```
perissology: there is already a static getter for type. Does that not work for instances? If not, we should at least delegate the call to `Campaign.type()` instead of adding `_type` to the instance

src/models/DAC.js
```diff
@@ -62,6 +63,10 @@ class DAC extends BasicModel {
     }
   }
 
+  get type() {
+    return this._type;
+  }
+
```
perissology: same comment as campaign

src/services/DonationService.js
```diff
@@ -8,6 +9,7 @@ import { feathersClient } from '../lib/feathersClient';
 import { getWeb3 } from '../lib/blockchain/getWeb3';
 
 import ErrorPopup from '../components/ErrorPopup';
+// import Campaign from '../models/Campaign';
```
perissology: remove please

src/contextProviders/DonationProvider.jsx
vojtechsimetka: I have a question here. This amount is the new `amountRemaining` once confirmed, right? In that case IMHO this is fine because both of these values need to be greater than 0, right? If anything I should actually be using the `pendingAmountRemaining` all the time rather than `amountRemaining`.

Issue #470, opened 2018-09-17
When donating to a milestone, the default suggested amount is blank


GriffGreen:
<img width="1050" alt="screen shot 2018-09-17 at 10 49 37 am" src="https://user-images.githubusercontent.com/14945613/45640462-99d3ea00-ba67-11e8-8c6d-4b73a8fe007c.png">

This is a different screen then when delegating from the my delegations screen... in that screen the entire requested amount is shown and you can use a slider to choose the amount... that screen works really well and makes it easy to donate as much as possible, it either caps out on the total you have to delegate, or the size of the milestone, whichever is less (the actual max) i would love to see this in the donation screen too!

<img width="986" alt="screen shot 2018-09-17 at 11 11 31 am" src="https://user-images.githubusercontent.com/14945613/45641627-d3f2bb00-ba6a-11e8-95d5-f33f6aba6937.png">

anshumanv:
Hey @vojtechsimetka @GriffGreen can I work on this?

vojtechsimetka:
Hey, sure!

vojtechsimetka:
This should work similar to delegate button.

anshumanv:
Ah missed this :sweat_smile: 

On it along with #474

anshumanv:
@vojtechsimetka can this be closed too or do we need some more changes here? :smile:

PR #482, opened 2018-09-27
Fixes #474: Do not allow negative values in the donation modal

### LINKED ISSUE #474: Do not allow negative donation values in donation modal
I think we should restrict the min amount to 0 rather than going in negatives.

image

anshumanv:
Fixes #474


vojtechsimetka:

Built with commit 98d87e0706de72aea611b8adbe286d33d2b7b20b

https://app.netlify.com/sites/giveth-dapp/deploys/5bac7e0ee39e7c026ad597f6

vojtechsimetka:

Built with commit 98d87e0706de72aea611b8adbe286d33d2b7b20b

https://app.netlify.com/sites/giveth-dapp/deploys/5bac7e0e82d3f12611d55ce1

anshumanv:
Though there's a catch, it limits to 0 only when you are reducing the value using the down key or using the down arrow, user can still type negative values, should I just remove the `-` sign as the user type or leave it as is. :thinking:

anshumanv:
Ah makes sense @vojtechsimetka.

One thing we can do is listen to onChange event on the box and if anything starts from `-` we just set the value to 0, what do you think of this approach? :thinking: 

Happy to hear your views.

vojtechsimetka:
> One thing we can do is listen to onChange event on the box and if anything starts from `-` we just set the value to 0, what do you think of this approach? 🤔

@anshumanv that sounds like a great solution. Could you extract it into a minimal component that would do so for the passed props min and max as well? It could be quite nice to have such component for other inputs as well. And it would be great if it could compare those props as bignumbers instead of just floats.

anshumanv:
Hey @vojtechsimetka after a lot of experimentation, Created a component and then I came up with 2 approaches

1. Using the existing Input component from `formsy-react-components`
There seem to be some issues with the onChange listener on this component and thus whenever I try to listen to a change event and set the state there's some lag and thus the input value isn't getting updated properly.

2. Using a normal HTML `<input />`
* No other characters allowed, values less than min are set to min.
* Not as good looking, validation messages won't show up.

What approach would you like @vojtechsimetka? :smile: Please let me know and I'll update the PR that way.

vojtechsimetka:
I don't actually like neither of these solutions :( . Looking at the [formsy input field ](https://github.com/twisty/formsy-react-components/blob/master/src/components/input.js)it looks quite complex, we definitely had issues with it as well and there was a plan to replace it. But for the time being I guess having the input lag is good enough. Please do option 1 and we will see how bad it is. Thank you @anshumanv !

anshumanv:
@vojtechsimetka did the approach you suggested, please see if this looks fine. :)

vojtechsimetka:
Hey @anshumanv that input does not work at all. Why did you decide to use another dependency? I thought we agreed to use the default input from formsy-react-components. Also as far as I can tell that component does not handle bignumbers nor does it handle the min amount (it only checks max).

anshumanv:
Hey @vojtechsimetka sorry I got confused a bit.

> I tried to use it, but it only lets me input 0.08 maximum amount (after pressing 1 twice).

Yes since the max prop passed to the component is the amount of ether in your wallet so it won't go above 0.08.

> Please use the Input field from react-formsy-components and in there check for props that are passed:

    min
    max

The intention is to limit the value in the donation field as per `min` and `max` right? I think this component already does this, can you please confirm? :smile: 

>  I thought we agreed to use the default input from formsy-react-components. 

The only problem is with the onChange listener with `react-formsy-component`'s `Input` component, it isn't getting updated properly as the value in the field changes, and we intent to limit the value in the field as per the `min` and `max` props passed right? So I need onChange listener on the field and as the user inputs the donation amount I just compare if the entered value is greater than max then the donation value is set to the `max` prop passed. Do you intent to use the prop in some other way? :thinking:

anshumanv:
> How about scratching all of this. Take the logic and everything from the delegation button and also complete the #470 issue at the same time. I can increase the bounty adequately.

Would be the best I think. :+1: :smile:  @vojtechsimetka 

Will solve it that way and complete the other issue as well. :+1: 

> The donate button now also does not work.

Also the problem I'm facing right now is about making test donations, how do I get the home ganache balance to test things out? Is it documented somewhere? :sweat_smile:

anshumanv:
Hey @vojtechsimetka added the slider to the donation modal component, now a slider appears while donating anywhere just like the delegation modal, please see. :smile:

vojtechsimetka:
Hey @anshumanv , thank you this is much better! I just have 2 edgecases and 1 optional improvement that I would really appreciate if you could solve:

1. When the metamask account is locked it looks very broken. I suggest not to display the inputs. 
screen shot 2018-10-17 at 23 31 18

2. Similarly, if your balance is 0 it is broken. The inputs should not be shown as well.
screen shot 2018-10-17 at 23 32 11

3. I would also appreciate one more thing (though I know it is not trivial). It would be great if when putting the slider all the way to the right the balance filled in all the small digits as well. Alternatively you could add a donate max button. E.g. using the addon on the input https://getbootstrap.com/docs/4.1/components/input-group/#multiple-addons

anshumanv:
Thanks for the review, on it!

vojtechsimetka:
Hi @anshumanv sorry to be pedantic but it seems you missed the 0 balance:
screen shot 2018-10-18 at 23 28 19

Also the button does not look properly styled and I suggest the slider to be below the input field - this way the * is correctly on the label.
screen shot 2018-10-18 at 23 29 01


Other than that this is great, I really appreciate it!

anshumanv:
Ahh, my bad.

Fixed the 0 balance issue (string comparison was needed :( )
screenshot from 2018-10-19 09-05-58

Also styled the donate-max button and fixed the asterisk position.

screenshot from 2018-10-19 09-15-32

Please see :smile:

PR #570, opened 2018-12-11
VideoPopup improvements

deamme:
Changes:
- Embedding Youtube videos should work now.
- Should preview the recorded video before upload

Things that I haven't been able to figure out:
- @GriffGreen is getting `null` when uploading via the camera option (I can't replicate this issue)


perissology:
okay, this is merged. not sure why github didn't recognize that

PR #632, opened 2019-01-19
Autofilling milestones

### LINKED ISSUE #631: Autofilling milestones
It would be great to be able to embed milestone information in the URL, meaning that a certain link would autofill certain fields when creating a milestone.

I talked to @Danibelle about using this for RewardDAO, but it could probably be useful for other things as well.

FrederikBolding:
Fixes #631

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "node-fetch": "^2.1.1",
     "node-sass-chokidar": "^0.0.3",
     "prop-types": "^15.5.10",
+    "query-string": "^6.2.0",
     "react": "16.4.1",
     "react-addons-shallow-compare": "^15.6.0",
     "react-avatar": "^2.3.0",
--- a/src/components/views/EditMilestone.jsx
+++ b/src/components/views/EditMilestone.jsx
@@ -7,6 +7,7 @@ import Toggle from 'react-toggle';
 import BigNumber from 'bignumber.js';
 import { Form, Input } from 'formsy-react-components';
 import GA from 'lib/GoogleAnalytics';
+import queryString from 'query-string';
 import Milestone from 'models/Milestone';
 import { feathersClient, feathersRest } from '../../lib/feathersClient';
 import Loader from '../Loader';
@@ -37,6 +38,18 @@ import MilestoneService from '../../services/MilestoneService';
 import CampaignService from '../../services/CampaignService';
 
 BigNumber.config({ DECIMAL_PLACES: 18 });
+
+const validQueryStringVariables = [
+  'title',
+  'recipientAddress',
+  'reviewerAddress',
+  'description',
+  'date',
+  'maxAmount',
+  'token',
+  'selectedFiatType',
+  'fiatAmount',
+];
 
 /**
  * Create or edit a Milestone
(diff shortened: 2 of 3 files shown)


FrederikBolding:
Everything seems to be working! Let me know if there's any issues!

package.json
```diff
@@ -28,6 +28,7 @@
     "node-fetch": "^2.1.1",
     "node-sass-chokidar": "^0.0.3",
     "prop-types": "^15.5.10",
+    "query-string": "^6.2.0",
```
vojtechsimetka: I don't think you need to add dependency. Just use the react-router param match function

src/components/views/EditMilestone.jsx
```diff
@@ -86,6 +100,7 @@ class EditMilestone extends Component {
   componentDidMount() {
     checkForeignNetwork(this.props.isForeignNetwork)
       .then(() => this.checkUser())
+      .then(() => this.props.getEthConversion(this.state.date, this.state.milestone.token.symbol))
```
vojtechsimetka: You are checking conversion before the symbol is set. This could potentially lead to incorrect exchange rate.

src/components/views/EditMilestone.jsx
```diff
@@ -1034,6 +1065,7 @@ class EditMilestone extends Component {
 
 EditMilestone.propTypes = {
   currentUser: PropTypes.instanceOf(User),
+  location: PropTypes.object.isRequired,
```
vojtechsimetka: I believe you don't need location to be passed, you can use react-router

package.json
vojtechsimetka: Unless this does something more but looking at the code you don't need it at all and it will actually be simpler

src/components/views/EditMilestone.jsx
```diff
@@ -128,13 +143,31 @@ class EditMilestone extends Component {
...
             .then(campaign => {
               if (campaign.projectId < 0) {
                 this.props.history.goBack();
               } else {
                 const { milestone } = this.state;
-                milestone.recipientAddress = this.props.currentUser.address;
+                validQueryStringVariables.forEach(variable => {
+                  if (qs[variable]) {
+                    if (variable === 'fiatAmount') {
+                      this.setMaxAmount('', qs[variable]);
```
vojtechsimetka: All these functions are doing setState. This is dangerous and should be avoided. I think it would be best to set the milestone object yourself and if needed at the end of this componentWillMount methods (but I don't think you need to)

src/components/views/EditMilestone.jsx
```diff
@@ -128,13 +145,46 @@ class EditMilestone extends Component {
...
+                      const token = React.whitelist.tokenWhitelist.find(
+                        t => t.symbol === qs[variable],
+                      );
+                      if (token) {
+                        milestone.token = token;
+                      }
+                    } else if (variable === 'date') {
+                      const date = moment(qs[variable])
+                        .utc()
+                        .add(1, 'hour');
```
vojtechsimetka: Why are you adding 1h?

src/components/views/EditMilestone.jsx
```diff
@@ -128,13 +145,46 @@ class EditMilestone extends Component {
...
             .then(campaign => {
               if (campaign.projectId < 0) {
                 this.props.history.goBack();
               } else {
                 const { milestone } = this.state;
-                milestone.recipientAddress = this.props.currentUser.address;
+                validQueryStringVariables.forEach(variable => {
+                  if (qs[variable]) {
+                    if (variable === 'fiatAmount') {
+                      milestone.fiatAmount = new BigNumber(qs[variable]);
```
vojtechsimetka: There should be a check for what was passed. If I pass do `fiatAmount=` it will throw and the website will just stay "Unable to load milestone".

vojtechsimetka: Same applies to other variables.

src/components/views/EditMilestone.jsx
vojtechsimetka: You should use getStartOfDayUTC from helpers

src/components/views/EditMilestone.jsx
vojtechsimetka: OK I can check that but I don't think adding it below and leaving it here makes sense.

FrederikBolding: Sure! Please do!

FrederikBolding:
@vojtechsimetka I fixed the comments you had, except for the exchange rate stuff, as you said you would look at it. Please take a look again.

PR #667, opened 2019-02-09
adding support for ERC712 pre-signed transactions

jvluso:


ewingrj:
So it looks like this was more of a POC. There's commented out code, and probably needs more work to be production ready.

I think we're going to go with the solution where you can use the dapp w/o having to signin. That should take care of the use case this is attempting to solve.

Thanks!

Issue #522, opened 2018-11-09
Need a way to save unfinished proposed milestones


GriffGreen:
... and campaigns... and dacs!

Right!

perissology:
for milestones at least, why don't you just put "WIP - " in the title and then propose it.

You can make the edits after proposing the milestone. When it's ready for the campaign owner to review, just remove the "WIP -".

For DACs & Campaigns, just flesh out the text in a text editor before creating. Once the text is ready, just create the dac/campaign. Then continue editing as needed.

To have a better way to do this in the dapp seems like very little reward for the work involved.

We could save the current progress to your browsers storage (like github does if you navigate away from a page after typing), but that would be at risk of loosing the data if you clear your browser cache for some reason

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0 DAI (100.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/Giveth/giveth-dapp/522/2418).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/Giveth/giveth-dapp/522/2418).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $54,966.53 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

eswarasai:
What's the proposed solution here? Let me know what exactly is needed to be done so that I can give it a try. Thanks!

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [charlie4biz](https://gitcoin.co/profile/charlie4biz) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/Giveth/giveth-dapp/522/2418?mutate_worker_action=approve&worker=charlie4biz) | [reject worker](https://gitcoin.co/issue/Giveth/giveth-dapp/522/2418?mutate_worker_action=reject&worker=charlie4biz))_.**

Will need more explanation though.

[...]

PR #689, opened 2019-02-25
Save Unfinished Milestones, Campaigns and DACs

e18r:
This addresses issue #522. A new file was created, Draft.jsx, which
defines a new button that allows to save drafts, as well as the draft saving functionality
itself. It was created with the objective of being reusable in different views. It is not a
component itself, but rather uses the context of the component that imports it. So far, the
saving functionality has been mostly implemented, as well as the event listeners that enable
and disable the button, and the button itself. The loading functionality is still missing,
and also there's a small bug that enables the button when it should be disabled.

There's a shortcoming to this approach that needs to be tested: LocalStorage only allows
for 5MB of storage per domain. In some browsers, this can be considerably less. This means
that large images will not be able to be saved, especially when having multiple drafts. A
failsafe has to be implemented for these cases.


GriffGreen:
Nice! Glad to see the quick progress @e18r :-D!

GriffGreen:
Looked at it! It looks great! 
Can i ask for the button to not show up at the begining (until the draft is save-able) so effectively there is 3 buttons... 

No button (before the draft can be saved... when there is nothing to save) 
Save Draft (before the draft has been saved)
Draft Saved (after the draft has been saved)

The other thing is that, will you be adding a button somewhere to continue the draft? Maybe in the campaign? or even in the header or something that is always saved? I'm curious on the best way to do that... I would have it sitting there in the header, AND next to propose milestone in the campaign... but maybe you have a better idea...  not sure how to easily manage the multiple milestone drafts edge case for the button at the top... and i don't want this bounty to get out of hand.

e18r:
Let's review all possible use case scenarios:

\ | New Milestone | Old Milestone
----- | ----------------- | --------------------
**New Draft** | **1**. User clicks on the "Propose Milestone" button and writes down the title: "Bla". Then her laptop runs out of battery. The title is automatically saved in the browser, but no milestone is created in the back end. | **3**. User selects the "Bla" milestone and clicks on the edit button. The title and the description appear in the form. She uploads a picture and clicks on "Save Draft". Then navigates away. The picture is saved in the browser, but nothing happens in the back end, so the "Bla" milestone still doesn't have a picture.
**Old Draft** | **2**. User clicks on the "Propose Milestone" button again. The word "Bla" is restored from the browser storage and appears in the title of the form. User writes a description and clicks on the "Create Milestone" button. The "Bla" milestone is created in the back end. | **4**. User selects the "Bla" milestone and clicks on the edit button. The title and the descrption appear in the form. The picture is restored from the browser storage and appears in the form. User clicks on "Update Milestone". The back end is updated, so the "Bla" milestone has now a picture.

If I understand you correctly, you propose a "Restore draft" button instead of it restoring automatically. In use case 4, this lets users decide if they want to edit the milestone as is, or edit the changes they had made in the past to that milestone. In use case 2, this lets users decide if they want to start a new milestone or continue editing a draft. In this case, the user could even save multiple drafts and later pick one from a dropdown list.

However, my suggestion is to restore drafts automatically and only for use cases 1 and 2. This would result in a simpler, more predictable feature that can be upgraded in the future if the need arises.

GriffGreen:
Sounds great! keeping it simple works!

Hoping the save draft button can hide until it can be pushed. And in your implimentation, pictures will be able to be saved still right? The rest sounds great @e18r :-D

SO SORRY FOR THE DELAY! Lost this tab yesterday.

e18r:
@GriffGreen great, let's keep it simple then. Also, the button will be hidden as you requested.

Pictures will be saved, yes, although only when they're smaller than 5MB. In some browsers, much less. In certain conditions, even (private browsing mode in some versions of some browsers with certain privacy settings), LocalStorage will be disabled.

You'll be hearing from me on Friday at latest.

e18r:
This one was hard work. It's almost ready. Only a few things missing:

* Blocking behavior
* Button state on form submit
* image loading
* user options loading
* attach to Campaigns and DACs
* automatic saving each minute

Please take a look @GriffGreen . Will be ready on Monday.

e18r:
It's taking longer than expected, sorry for that. EditMilestone.jsx had a lot more complexity than what I had thought. I'm working on the toggles it has and also I will be working on the option to add "line items". I'll keep working on Friday.

GriffGreen:
Hey @e18r do you consider this to be done?

e18r:
Hi @GriffGreen not quite yet. There is still some work needed on EditMilestone.jsx. Also, should I set an automatic draft saving each minute? Last time we talked about this feature, you seemed doubtful about it.

e18r:
The feature is almost completed. On Friday I will make the draft button responsive, and that will be it.

GriffGreen:
For the 1 minute thing, its all gravy as long as we can easily adjust it if we notice any issues.

GriffGreen:
@e18r there is a dapp meeting at 6pm Barcelona time if you had any questions about this :-)
https://meet.jit.si/giveth-dapp

e18r:
I finished! I made the button responsive and set an auto save each minute. I also solved a bug I had introduced.

Please review it carefully and let me know if you find anything unexpected.

ElMejorge:
Hello @e18r, while doing QA on the feature I found that if you move around after creating any resource (milestone, campaign and DAC) then the draft is not erased and it gets reloaded when you try to create another one. This makes it so that you have to erase everything since you already created the resource the draft is from and there's no way to delete it.

I think for this it would be better to delete the draft after the transaction is made instead of waiting for the transaction to be validated as I think is happening now. 

Tell me if you have any trouble replicating this.

e18r:
Hi @ElMejorge, I did as you requested. Now drafts are deleted immediately after submit, without waiting for the milestones, campaigns and DACs to be mined.

PR #742, opened 2019-04-06
Closes #679 - Right click Milestones to open in new tab

### LINKED ISSUE #679: Right click Milestones to open in new tab
Unicorn DAC has me scanning the campaigns for milestones that meet funding prioritization criteria, and having to 'go back' to the full list every time I select one to review is a real pain as the lists get longer.

From the main page it would be AWESOME to right click on a DAC, Campaign, or Milestone and open several in new tabs.

gil-air-may:
Closes #679 

To solve this, I switched one of the divs in `DacCard.jsx` with an `<a/>`. Then I added an appropriate `href` to that element. Now that we have a link, you can right-click -> "Open in new tab" 👍  

I also had to change the `_card.scss` styles in order to prevent the default `<a/>` styling from appearing.

### DIFF
--- a/src/components/CampaignCard.jsx
+++ b/src/components/CampaignCard.jsx
@@ -17,10 +17,15 @@ class CampaignCard extends Component {
     super(props);
 
     this.viewCampaign = this.viewCampaign.bind(this);
+    this.createCampaignLink = this.createCampaignLink.bind(this);
   }
 
   viewCampaign() {
     history.push(`/campaigns/${this.props.campaign.id}`);
+  }
+
+  createCampaignLink() {
+    return `/campaigns/${this.props.campaign.id}`;
   }
 
   render() {
@@ -30,12 +35,11 @@ class CampaignCard extends Component {
       <div
         className="card overview-card"
         id={campaign.id} // eslint-disable-line no-underscore-dangle
-        onClick={this.viewCampaign}
         onKeyPress={this.viewCampaign}
         role="button"
         tabIndex="0"
       >
-        <div className="card-body">
+        <a className="card-body" href={this.createCampaignLink()}>
           <div className="card-img" style={{ backgroundImage: `url(${campaign.image})` }} />
 
           <div className="card-content">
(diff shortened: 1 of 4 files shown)


ElMejorge:
Hello @gil-air-may, great of you to contribute like this, before I merge could you do the same for the other cards (milestones and campaigns)?

gil-air-may:
Sure @ElMejorge! I will do the same for the other cards.

gil-air-may:
Now, you can right click any card and open them on a new tab. 

For the campaign cards, I used the exact same logic on `src/components/CampaignCard.jsx`

campaign-right-click

However, for the milestone card I had to take a slightly different approach.

Notice how to milestone card contains a `<div className=card-avatar />` which allows you to check the creator's profile. It also has a button which allows you to edit the milestone. Both of these elements use `onClick` attributes. This header used to be inside `<div className=card-body />`. In other words we had something like:

```
<div className=card-body>
    <div 
        className=card-avatar 
        onClick={this.viewProfile}>
        <button onClick={e => this.editMilestone(e)}/>
    </div>
</div>
```

If I simply changed the outer element to an `<a/>` the link behavior would always override both onClick behaviors. Basically, you would always be redirected to the milestone page if you clicked any of the buttons on the `card-avatar`.

So to solve this I had to rearrange the elements a bit. Here's the new jsx structure for `src/components/MilestoneCard.jsx`:

```
<div className=card-body>
    <div 
        className=card-avatar 
        onClick={this.viewProfile}>
        <button onClick={e => this.editMilestone(e)}/>
    </div>
</div>
<a className=card-body>
    [rest of the divs on the card body]
</a>
```
I created a dummy outer element for the card avatar, to maintain inherited styles. So for the milestone card, you can right click to open in new tab, as long as you don't do it inside the `card-avatar` (top) section of the card. Check it out:

milestone-right-click-1

This way, the onClick behavior of the `card-avatar` is preserved :smiley:

milestone-right-click-2gif

ElMejorge:
Great job and thanks for the thorough explanations, they really help with the QA process. I'll go ahead and merge and close the issue.

PR #763, opened 2019-05-14
Update Web3Provider.jsx

GriffGreen:
Could have just committed, but figured a PR is fun


GriffGreen:
wow epic fail... we need to merge pull requests faster than this... its clearly not even being looked at

Issue #766, opened 2019-05-15
Fixing Travis Integration so contributors stop wasting their time 


GriffGreen:
https://github.com/Giveth/giveth-dapp/pull/757#issuecomment-489735406
This is an issue... we have travis doing checks but the checks fail... we need to better maintain this. We are wasting every contributors time when this check fails :-(

nadirlaskar:
It seems there is some problem with eslint check in the  build [process](https://travis-ci.org/Giveth/giveth-dapp/builds/528971192?utm_source=github_status&utm_medium=notification)

```
20.92s$ npm run eslint
> giveth-dapp@0.1.0 eslint /home/travis/build/Giveth/giveth-dapp
> eslint 'src/**/*.js*'
/home/travis/build/Giveth/giveth-dapp/src/components/ArchiveMilestoneButton.jsx
  87:9  warning  Unexpected console statement  no-console
/home/travis/build/Giveth/giveth-dapp/src/components/BackupWalletButton.jsx
  20:15  error  Replace `⏎······Array.isArray(wallet.keystores)·&&⏎·····` with `·Array.isArray(wallet.keystores)·&&`  prettier/prettier
  23:7   error  Delete `··`                                                                                           prettier/prettier
  24:1   error  Replace `··········` with `········`                                                                  prettier/prettier
  25:9   error  Delete `··`                                                                                           prettier/prettier
  26:1   error  Replace `··········` with `········`  
```

Some components are having inconsistent spaces i believe, which is failing the eslint check

GriffGreen:
@nadirlaskar how much would you charge us to fix this?

nadirlaskar:
> @nadirlaskar how much would you charge us to fix this?

There are 45 files having lint issue in the repo right now, some files are fixable with eslint --fix, others require manual changes.

Some files are using label without `htmlFor` attribute, fix requires updating the property where possible or replace with `<span className="label"></span>`

Also some update is required in `eslint` & `eslint:fix` script in `package.json`

All changes are possible to be completed in 2 hrs,
I can start working on a pull request with the described fixes for 60 DAI

nadirlaskar:
> @nadirlaskar how much would you charge us to fix this?

I have opened a pull request with a fix, check it out @GriffGreen

GriffGreen:
NICE! Whats your address? we are making a milestone for you in the dapp ;-D

nadirlaskar:
[...]

PR #773, opened 2019-05-19
FIXES: Travis Integration, ESLint and Test cases, problems

nadirlaskar:
#766 

There are 45 files having lint issue in the repo right now,

Some files are using label without `htmlFor` attribute, fix requires updating the property where possible or replace with `<span className="label"></span>`

Also quotes are required to be removed from  `eslint` & `eslint:fix` script in `package.json` for it to work in windows environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,8 +82,8 @@
     "build": "npm run build-css && react-scripts build",
     "serve": "npm run build && pm2 serve -s build",
     "test": "react-scripts test --env=jsdom",
-    "eslint": "eslint 'src/**/*.js*'",
-    "eslint:fix": "npm run eslint --fix",
+    "eslint": "eslint src/**/*.js*",
+    "eslint:fix": "eslint --fix src/**/*.js*",
     "eject": "react-scripts eject",
     "prettify": "prettier --write 'src/**/*.js*'",
     "doc": "jsdoc src -r -d docs"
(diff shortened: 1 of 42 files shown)


nadirlaskar:
Following test case are inconsistent with models,

1. Test case for User model is checking `commitTime` attribute type as Number but test case is using `commitTime` as string in the test data.
     > _current-fix: updated test data to number from string._

2. Donations model have remainingAmount property, which is missing in test data.
     > _current-fix: updated model to use amount when remainingAmount is absent._
     `new BigNumber(utils.fromWei(data.amountRemaining?data.amountRemaining:data.amount));`

3. Donations model is converting `wei` to `eth` for amount and remaining amount attribute, which breaks the test case check for equality. 
     > _current-fix: updated test case data to check in eth_

nadirlaskar:
@ElMejorge did you get a chance to review the changes?

nadirlaskar:
@GriffGreen Linking my [eth address](https://github.com/Giveth/giveth-dapp/issues/766#issuecomment-494065941) comment, excited for my first [bounty](https://github.com/Giveth/giveth-dapp/issues/766#issuecomment-493718928) hunt 👍 

> ETH Address `0x039dACca4a447E140d8B29F0652cd844Ea827D1a`
> Bounty `60 DAI`

Thanks :-)

PR #878, opened 2019-07-07
Fixes #769

### LINKED ISSUE #769: Everytime i paste when making a campaign i get scorlled to the top of the window
i hit ctrl-v and then i scroll out of view from where i was pasting... its pretty lame!

aminlatifi:
Fixes scroll problem at the end of react-quill component

### DIFF
--- a/src/components/QuillFormsy.jsx
+++ b/src/components/QuillFormsy.jsx
@@ -182,6 +182,7 @@ class QuillFormsy extends Component {
           placeholder={placeholder}
           onChange={setValue}
           id="quill-formsy"
+          scrollingContainer={document.documentElement}
         />
         <span className="help-block validation-message">{errorMessage}</span>
       </div>


aminlatifi:
> It's a lot better now!, an optimal solution could be scrolling to the bottom of the quill, since when you paste big text it still leaves you where you started and you need to scroll again unfortunately, do you think that would be possible @aminlatifi?

I feel it could not be solution, because user may paste in the middle of text. Scrolling to the bottom does not make sense in this case.

arbreton:
@aminlatifi agreed, we need to review this, optimal solution should be improving the UX completely so we don't have to come back to this in the future

Another optimal alternative would be **limiting the text box size** just like this comment box on GitHub, right now the box sizes infinitely depending on the text, when it can just scroll with a limited size and if you paste something you are always seeing the last thing you pasted.

Do you think that would be viable?

aminlatifi:
@arbreton It makes sense. But I don't know scrolling to the bottom of quill is still good idea or not? 
However, this comment box on github is not limited by size. 
I think we should first implements our ideas and see the outcome. I also thought about saving scroll position and restoring it, but it would have other problems.
If our ideas are not promising, maybe we should find another element than quill. What do you think?

arbreton:
That sounds great!, yeah sorry about that, not scrolling to the bottom of the quill that will not work,
actually the comment box on github is very smart, it limits the max size to the bottom of the screen, i think maybe we can gather some of this ideas and implement them.

What i would do is get this changes merged and i will open a thread so we don't lose this suggestions

PR #888, opened 2019-07-11
Report an issue with error info using error popup

### LINKED ISSUE #872: Error messages should link to https://github.com/Giveth/giveth-dapp/issues/new
Let's get more issues, and faster customer service!

We are about to have a pretty big beta test with Burning man and a lot of people will end up getting milestones sent to them... if we add a link to https://github.com/Giveth/giveth-dapp/issues/new in every error message, we will drastically reduce the barrier to finding out what is broken!

aminlatifi:
Fixes #872

### DIFF
--- a/src/components/ErrorPopup.jsx
+++ b/src/components/ErrorPopup.jsx
@@ -53,12 +53,6 @@ export default (shortDescription, error) => {
       window.open(
         `mailto:${config.bugsEmail}?subject=Error in DApp&body=${encodeURIComponent(body)}`,
       );
-    } else if (value === 'gmail') {
-      window.open(
-        `https://mail.google.com/mail/u/0/?view=cm&fs=1&tf=1&to=${
-          config.bugsEmail
-        }&su=Error in DApp&body=${encodeURIComponent(body)}`,
-      );
     } else if (value === 'github') {
       window.open(`${config.githubUrl}/issues/new?body=${encodeURIComponent(body)}`);
     }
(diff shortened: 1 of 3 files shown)


arbreton:
I have checked the implementation, is excellent Amin!, it will be really convenient for the users of the dapp,
Only thing now is i think we need to figure out what to do with the other buttons since now we have a direct reporting tool to GitHub i don't know if we even need the other buttons because they can be quite confusing, i think **we can just leave the report issue** which will redirect you to github issues, we may need some external feedback maybe from @GriffGreen, or @markoprljic regarding #872 

Screen Shot 2019-07-11 at 5 55 42 PM

aminlatifi:
Thanks,
I agree, it seems other button are excess and the dialog will be neater after removing  those buttons.
I also searched to find a way to show github icon, but it's not possible by this library.

markoprljic:
@arbreton If the Report button redirects to Github does it automatically opens blank page for adding new issue? Is there a way to prepopulate github issue title with default text such as "Report an Issue with Giveth dApp" or something like that?

aminlatifi:
> @arbreton If the Report button redirects to Github does it automatically opens blank page for adding new issue? Is there a way to prepopulate github issue title with default text such as "Report an Issue with Giveth dApp" or something like that?

I feel it should be blank, in order to force submitter to fill out appropriate title manually. 
Nevertheless, title can be filled with some default value, provided someone else be responsible to edit such issues with default title. However I don't know issue title can be changed by someone other than issuer.

arbreton:
New comments @ #872

PR #897, opened 2019-07-13
Feature 866 uncapped milestone title helper

aminlatifi:
Helper title added to milestone which tells which tokens are acceptable.

### DIFF
--- a/src/styles/_sweet-alert-overrides.scss
+++ b/src/styles/_sweet-alert-overrides.scss
@@ -12,6 +12,10 @@
 
 }
 
+.swal-wide {
+  width: 550px;
+}
+
 .swal-text {
   text-align: center;
 }
--- a/src/components/ErrorPopup.jsx
+++ b/src/components/ErrorPopup.jsx
@@ -11,30 +11,42 @@ export default (shortDescription, error) => {
   const errorHandler = value => {
     let body;
     if (error instanceof Error) {
-      body = `
-      Description of the Error:
-      ${shortDescription}
+      if (value !== 'github') {
+        body = `
+        Description of the Error:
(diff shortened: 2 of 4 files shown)


src/styles/_sweet-alert-overrides.scss
```diff
@@ -12,6 +12,10 @@
 
 }
 
+.swal-wide {
```
arbreton: What is this for?

aminlatifi: This was for other issue related to #872 
As I remember, I create each branch from develop. I don't know why this line exists in scss. Howere, this should not be applied in this commit. Thanks @arbreton .

aminlatifi:
> Hello Amin, i checked the title looks great, is going to be very useful!,
> Few details:
> All the changes from Github reporting please leave them out on a separate branch, **always use develop branch** instead of your last branch **as base**, this will prevent changes that haven't been approved to make it into develop, that is due to the Github reporting feature is still being discussed
> 
> The title helper should appear as well on capped and uncapped milestones to keep consistency on both, right now is not showing up on milestones accepting only DAI for example http://localhost:3010/campaigns/5b39d45e14cec916d00dab20/milestones/5d246e7f67f1fe662d8ced62 using beta environment
> 
> A little insight on the scss modification
> 
> Also when creating your branch you can try using a more simple format so it can be easier to read and for the record to keep track, we are using **b for bugs**, **f for features**, something like **f_866_title_helper**
> 
> Keep me posted!, excited to see this one released

I cannot visit the link since it is on your localhost. Title helper has some output on every condition, so it should not be empty in logic.

Issue #342, opened 2018-05-04
Show # of confirmations for pending dac/campaigns/milestone, etc


ewingrj:
For all of the following views

MyDacs
MyCampaigns
MyMilestones
MyDelegations
MyDonations

If the status is pending, we should show the current # of confirmations and the # of required confirmations. This data should all be present in the return models from feathers.

cbrenn:
Working on this issue now. 

Could you explain conceptually what a confirmation is? 

I think that it is a reviewer must confirm the campaign is valid before it can be launched and this confirmation takes place on the reviewer page. 

I'm thinking about making it a property of the campaign model, but I don't see it stored anywhere currently.

perissology:
a confirmation is just the number of blocks passed when the tx was mined. If it was mined 2 blocks ago, then there are 2 confirmations.

This data is already present on the model returned from feathers, you only need to show it in the ui

vojtechsimetka:
Hi @cbrenn, thank you for your interest. See perissology response. Will you take this up? If so let me know. I was just about to assign this to @xavivives

cbrenn:
Coming along with this, and should be finished by Sunday. The only place I really see a clear confirmation state is in the Donation Model. Everything else I am assuming I take from the model directly, unsure about changing the constructor params for campaign, milestones etc. 
Please let me know if that is ok.

vojtechsimetka:
Please go ahead and modify the other models. The required data should be received from feathers, but they are not added to the other models.

cbrenn:
Confirmations seem to be attached to the idea of a transaction. Milestones and delegations don't seem to be mapped to the idea of a transaction. Also they don't have a model in the D'app and therefore don't take in confirmations or required confirmations as state. Not seeing a logical way to get transactions from milestones or delegations...something I'm missing?

perissology:
every action taken on a milestone (mark complete, reject, etc) is a tx and the txhash & # of confirmations on the milestone is updated.

For delegations, the txhash & # of confirmations is updated on the donation that is delegated

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0 DAI (100.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/Giveth/giveth-dapp/342/2419).
[...]

PR #924, opened 2019-08-01
Number of confirmations are updated for pending items.

aminlatifi:
Refs #342
Number of confirmations are now updated on the UI. Feathers does not propagate changes on this field because #  confirmations is added by a hook (addConfirmations) in feathers. Actually, what makes change in #  confirmations are changes in corresponding event, and events are not exposed by feathers right now. 
Therefore, I decided to poll just pending items to update UI if their # confirmations is changed. Not the best way, but the best we can do without changing feathers so far.
Implemented for:
- Campaigns
- DACs
- Milestones

Number of confirmations are not available for MyDonations and MyDelegations, featherjs does not provide such information till the item is committed.

In order to meet #342 decently we need to do modify feathers and after take a better action on DAPP

### DIFF
--- a/src/components/views/MyCampaigns.jsx
+++ b/src/components/views/MyCampaigns.jsx
@@ -40,6 +40,27 @@ class MyCampaigns extends Component {
   }
 
   componentDidMount() {
+    this.updatePending = setInterval(() => {
+      if (this.isPendingCampaign) {
+        this.state.campaigns.data
+          .filter(c => c.confirmations !== c.requiredConfirmations)
+          .forEach(c => {
+            const currentCampaign = c;
+            CampaignService.get(c._id).then(newCampaign => {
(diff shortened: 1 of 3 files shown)


aminlatifi:
> Confirmations should hide when the process is done

It was not issue requirement but good idea. I will implement this if review is needed.

> Also confirmations is referring to block confirmations on the blockchain!,

So if it should not be referring to this, what else it should be. The issue was that # confirmations are not updated on UI. Nothing else is mentioned about confirmations concept.

> It is strictly necessary to modify feather if needed to get the transaction blockHeight vs the currentBlockHeight

I agree. This needs redesign some part of feathers and tables (documents) on DB. We should discuss it specifically on other conversation.

arbreton:
We will resume progress at later time

PR #956, opened 2019-08-15
Paid milestone details

### LINKED ISSUE #604: Date of Donations on Milestones has a confusing UI
When I look at a milestone before it's reached "Paid" it displays the different dates that donations were made. 
screen shot 2018-12-29 at 1 11 32 am

But then, when it's paid out, it shows only the date that all were paid out, next to each milestone as if that was the date of donation.

screen shot 2018-12-29 at 1 11 15 am

I'm guessing it's showing the date of the most recent status change on the donation. At least this is how it all appears to me. 
But it would be way better to just keep the date of donation/delegation right there, it helps differentiate.

### LINKED ISSUE #873: donations are combined once they are paid out
Screen Shot 2019-07-06 at 12 19 01 AM

Funny thing... i'm pretty sure there is another issue that Lorelei made a while ago about how the donations update at the bottom of a milestone... it aggregates the multiple donations i sent there and combined them to one, changing the date to the payment date instead of the donation date... 

Its seems like payments should be a different heading and we shouldn't lose the data about who donated when.

aminlatifi:
Fixes #604 
Fixes #873

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -22,14 +22,10 @@
     -->
 
   <!-- Bootstrap 4 -->
-  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/css/bootstrap.min.css" integrity="sha384-/Y6pD6FV/Vv2HJnA6t+vslU6fwYXjCFtcEpHbNJ0lyAFsXTsjBbfaDjzALeQsN6M"
-    crossorigin="anonymous">
-  <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN"
-    crossorigin="anonymous"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.11.0/umd/popper.min.js" integrity="sha384-b/U6ypiBEHpOf/4+1nzFpr53nxSS+GLCkfwBdFNTxtclqqenISfwAzpKaMNFNmj4"
-    crossorigin="anonymous"></script>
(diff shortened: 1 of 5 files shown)


public/index.html
```diff
@@ -22,14 +22,10 @@
...
   <!-- Bootstrap 4 -->
-  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/css/bootstrap.min.css" integrity="sha384-/Y6pD6FV/Vv2HJnA6t+vslU6fwYXjCFtcEpHbNJ0lyAFsXTsjBbfaDjzALeQsN6M"
-    crossorigin="anonymous">
-  <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN"
-    crossorigin="anonymous"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.11.0/umd/popper.min.js" integrity="sha384-b/U6ypiBEHpOf/4+1nzFpr53nxSS+GLCkfwBdFNTxtclqqenISfwAzpKaMNFNmj4"
-    crossorigin="anonymous"></script>
-  <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/js/bootstrap.min.js" integrity="sha384-h0AbiXch4ZDo7tp9hKZ4TsHbi047NrKGLO3SEJAg45jXxnGIfYzk4Si90RDIqNm1"
-    crossorigin="anonymous"></script>
+  <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
```
arbreton: I think this made it through the pull request as a mistake

aminlatifi: No, bootstrap version is upgraded intentionally. Actually, some components added does not work with beta version.

aminlatifi:
> I love the idea of the nested list!, it's so great, you can report them as finished by now, i think we are just in wait for the colors, could be better to leave them white to keep consistency, we should wait to ask for a second opinion, but looks amazing

It's good that functionality is acceptable. I used bootstrap secondary colors, of course better colors exists. 
If you reached to any better decision on colors please let me know.

PR #1025, opened 2019-10-11
Add deploy funds forwarder factory button

dapplion:
Calls 
```
fundsForwarderFactory.newFundsForwarder(giverId, receiverId)
```
where
```
receiverId = campaign.projectId
giverId = (campaign._owner || {}).giverId
```
**NOTE**: The contract is not deployed yet and its address is a placeholder. Metamask should you warn you about it too.

----

#### Questions:
- Everyone should be able to see the button or only the campaign owner?
- Better naming?

----
#### General view

generak

----

#### Modal view after button click

modall


arbreton:
I think we will just need maybe an opinion from @GriffGreen regarding permissions or naming and also we might need the address before release

GriffGreen:
Anyone should be able to deploy.

What if the button just says, "Create Donation Address"

I'm curious to see how the collection process works :-D

Great work DAppLion :-D This will make it sooooo easy for DAOs to work with the DApp!!

dapplion:
Deployed the FundsForwarderFactory supporting MolochDAO forwards to mainet https://etherscan.io/address/0xa04d4f31c2aaceae21279f4c1143741625d79efb#code

I have added a commit on this PR with:
- Mainnet address of the contract in the config
- Fetch the resulting address from the tx logs and show it in a react.toast

**However** I don't think this functionality is safe to be exposed to regular users:
 - Right now the user can't tell if there already exists a `fundsForwarder` deployed contract for this campaign, so he/she may end up deploying the contract over and over again
- Aside from the button how would the already deployed tx be shown to other users or to that same user?
- How to prevent re-deploys while the transaction is being mined? Maybe cache the fact that a contract is being deployed in feathers or in local storage?

Finally, the client to call `forward` funds is yet to be developed. It should first figure if there is a fundsForwarder for that address and then query the balances and if there are enough balances offer somewhere to call one transaction (or multiple depending if tokens + dao or only tokens).

These functionalities require past event querying, @arbreton  should it be done by feathers, or hardcore by the react component on each load?

arbreton:
@dapplion i think the best way i can think of is to use feathers since we already have it and take advantage of this system to cache this kind of data

If the FF is already deployed after the address needs to be saved to feathers so it saves it into the campaign object in MongoDB that way we will know if a FF has been already been deployed before

I can suggest if a FF is already deployed the button can change from "Create donation address" to "View donation address", which the "View" will be available to the public, and "Create" only the owner can do this action

While the Tx is pending we can definitely use cache to check on this Tx status if it is still processing 

For the past event querying we already have similar implementations on feathers, i completely agree we can use that for that scenario

Please let me know what you think

dapplion:
@arbreton I agree it's the most cohesive approach given the current architecture. Can you take care of the feathers part? I have never opened that codebase. If so let me know if you need help understanding how to fetch the required information.

bowensanders:
I posted this to Riot, but figured it should also go here:

> Update on direct donation to a campaign from an Aragon DAO on Mainnet
I successfully created a new (Mainnet) Aragon DAO
I created a donation address via the button on the Gov Campaign - https://etherscan.io/tx/0x0720331fa727174729339318b12cbec0f1cd0273b014dfac2f6231e77bec29ae
I used Frame / Agent via beta.mycrypto.com to execute a vote to donate 0.001 ETH directly to that new address - https://etherscan.io/tx/0xa049c1b7a48a466cdede54879300cdd0f7f2b54df38a48494360803e635aeabb
The donation now appears to be sitting on the created donation address - https://etherscan.io/address/0xe7a6630d8ce00b6b044dc7c00e7f294ed97a6dfe

Not sure if there are any more steps to be done before the donation makes its way to Giveth - but all the steps completed successfully! I have screenshots of the process if necessary.

arbreton:
@bowensanders nice!!, thats awesome!, i think @dapplion already replied on Riot, we still managing the details 

Speaking about it @dapplion , yes of course, i can make this modifications on feathers for the cache on the DB, i will modify a version of this branch and update the PR with the details

WIP

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

dapplion:
@bowensanders @arbreton I have developed the functionality to view balances and forward on the button itself. Ideally what I have done should be re-organized somewhere else but now the full UX can be completed.

Please, @bowensanders feel free to finish your test now. I attach screen capture below of how it should look like

----

viewdon

----

forwardbal

arbreton:
@dapplion after @bowensanders testing on production what happened was that the forwardFunds to the bridge made 1 tx with 2 donate events because there was 2 currencies that needed forwarding.

As per the current implementation of the bridge monitor, the cache DB uses unique constraint format, and relates 1 tx to 1 event, otherwise it crashes the service

Would you think you would be able to modify the funds forwarder smart contract to make 1 tx per currency so it generates 1 tx per event?

As i think modifying the bridge monitor code could result on critical loss of sight on important information

What do you think @dapplion ?

dapplion:
> Would you think you would be able to modify the funds forwarder smart contract to make 1 tx per currency so it generates 1 tx per event?

There's no need to modify the smart contract, the client should use the single currency forwarder function `forward` instead of `forwardMultiple`. 

> As per the current implementation of the bridge monitor, the cache DB uses unique constraint format, and relates 1 tx to 1 event, otherwise it crashes the service

If it really crashes the service you should work on a fix asap in parallel so you can phase the new monitor to production without downtime.

GriffGreen:
ok so then we need to change the UI to add a second button... one that says to Forward DAI and one that says to Forward ETH and then we make it ready for multiple token support. (Can;t wait for rDAI!!)

dapplion:
Changed the code to only call `forward(address)`. To do so a button is showed for each approved token in a generic fashion. To add new tokens one should just declare their address and symbol in the array. 

forward-one-by-one

*Note that balances are fake; to allow working on the UI :)*

GriffGreen:
And this will work with the new dai right?

dapplion:
> And this will work with the new dai right?

The smart contracts are agnostic to what token is used. The only requirement is that the bridge allows it.

For the front-end, the only change necessary would be to add the correct contract addresses and symbols to an array in front-end config. What naming convention do you want to follow?

PR #1047, opened 2019-11-09
Donation collect count limit moved to config file

### LINKED ISSUE #1040: I was warned about doing multiple transactions but the message implied i should do 1
Screen Shot 2019-10-21 at 11 40 23 AM

https://beta.giveth.io/campaigns/5d2602a167f1fe662d8cee2a/milestones/5da93898d25d7f4a295e3d47

I think the limit is still set to 5 but it should go back to 7... I wouldnt change the text in the warning... i would change the limit back to 7

aminlatifi:
* Fixes #1040

### DIFF
--- a/src/components/WithdrawMilestoneFundsButton.jsx
+++ b/src/components/WithdrawMilestoneFundsButton.jsx
@@ -11,6 +11,7 @@ import { checkBalance, isLoggedIn } from 'lib/middleware';
 import { Consumer as Web3Consumer } from '../contextProviders/Web3Provider';
 import DonationService from '../services/DonationService';
 import LPMilestone from '../models/LPMilestone';
+import config from '../configuration';
 
 class WithdrawMilestoneFundsButton extends Component {
   async withdraw() {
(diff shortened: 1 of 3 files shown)


src/components/WithdrawMilestoneFundsButton.jsx
```diff
@@ -32,11 +33,12 @@ class WithdrawMilestoneFundsButton extends Component {
           title: isRecipient ? 'Withdrawal Funds to Wallet' : 'Disburse Funds to Recipient',
           content: React.swal.msg(
             <div>
-              {donationsCount > 5 && (
+              {donationsCount > config.donationCollectCountLimit && (
```
arbreton: This might need to be replaced with ">=" so the config.donationCollectCountLimit matches the text below

aminlatifi:
> Everything looks great, just an important note to adding this implementation, there is one other place that this variable is being used, it is DonationService.jsx: Line 805, $limit: is directly related to this and needs to be changed at the same time as this variable
> 
> Also after some investigation the max limit can be returned to 15 on the configuration

If you mean the variable (5 now) should be changed by config variable (donationCollectCountLimit), it has already done. If not, please explain more because I didn't get that.

src/components/WithdrawMilestoneFundsButton.jsx
aminlatifi: If donatoinsCount === config.donationCollectCountLimit , all donations can be collected in one transaction. So `>` seems to be correct.

Issue #1031, opened 2019-10-15
"Use My Address" does not work after once manually edited


aminlatifi:
![Use My Address problem 2019-10-15 15_13_51](https://user-images.githubusercontent.com/5684607/66828825-db7edf00-ef5e-11e9-9f14-f64d6f1c3bc0.gif)

Issue #991, opened 2019-09-20
toggle buttons on milestones are very slow


lorelei1:
The "Give 3 percent to a DAC" toggle and the "Disable Reviewer" toggle both take a looooong time to register that you have clicked them... we're talking over a minute

arbreton:
This is a known issue @lorelei1!, actually they seem to stop responding due to conflicting cache, try deleting your browser cookies and they should be working properly again, investigating the root cause

lorelei1:
it fixed it temporarily... but it got slow again pretty soon! after one or 2 more milestones

PR #1055, opened 2019-11-17
H 995 fiat amount update problem

### LINKED ISSUE #995: Old bug! Still goin
If you don't click which currency you're raising funds in before inputting the amount - the conversion gets stuck and you end up with something nonsensical like this: 
Screen Shot 2019-09-20 at 4 06 43 PM
Raising 3500 DAI, aka 16 DAI ... 🤦‍♀️

aminlatifi:
* Fixes #995, #1031, #991


aminlatifi:
Important refactoring in EditMilestone view which will solve several high appearing issues.

PR #1063, opened 2019-11-24
Feature/portis support

mimafogeus2:


aminlatifi:
@mimafogeus2 Great work! Thanks for adding Portis support.
I tested that, it seems OK almost. Just I couldn't find a way to switch between networks to do donation. Hopefully, Portis has a [method](https://docs.portis.io/#/methods?id=changenetwork) for this requirement in its API. @mimafogeus2 can you address switching between networks too?

I pushed your job to rep in a new branch https://github.com/Giveth/giveth-dapp/tree/wallets_support.
We need to support Frame wallet too, and I think this branch is a good place to prevent conflicts. Please create other pull requests to this branch.

mimafogeus2:
Hi,

Thanks for your prompt response!

Portis' SDK allows you to change network. However, there's no UI for this,
so we'll have to implement this in Giveth.
What networks are supported?

One more thing - I've seen that Netlify branch deployments are broken. This
is due to each one having a custom subdomain that belongs to Netlify, which
is impossible to safely whitelist.
Netlify supports custom domains, and using such a domain will solve this
problem, and any future one where a third party service needs domain
whitelisting.
Is it possible to do this?

aminlatifi:
@mimafogeus2 Networks can be found in config

Regarding Netlify deployment error, I created a pull request from wallets_support branch to develop just to see how it goes. There is problem in Netlify deployment, you're right. 
I chased the root cause and I found Portis library makes problems for our tests (jest). It seems we have nothing to do with that, so I reported the problem as an issue here: https://github.com/portis-project/web-sdk/issues/50
We could continue our development on wallets_support branch, test it on our local machines and hope portis team solve the problem in mean time.

aminlatifi:
And it's the link to pull request: https://github.com/Giveth/giveth-dapp/pull/1065
@mimafogeus2 please apply all your changes to wallets_support branch

PR #1232, opened 2020-04-27
Fixed bugs of re-propose milestone and reject completion request

aminlatifi:


aminlatifi:
rePropose message context is supported on milestone conversation.
Reject Completion Request needs user to be signed in.

Issue #1339, opened 2020-07-06
Aggregate the leader board so that it sorts by the top donors not the top donations  


GriffGreen:
<img width="1012" alt="Screen Shot 2020-07-06 at 2 27 51 PM" src="https://user-images.githubusercontent.com/14945613/86593137-fda90900-bf94-11ea-93c4-1a9e7a9ec1d8.png">
https://beta.giveth.io/campaigns/5d110631994eac53adc8f21e

One donation of 25k is on top but there are several donors that have donated more than that... Santa and Anonymous Donors (or Gitcoin)

GriffGreen:
Make this a leader board and not a list of the Donations!

List the USD Value, using the date that the ETH hits the milestone or campaign

aminlatifi:
Need to redesign the donation table... we should aggregate based usdValue of donations

krrisis:
okay so this was supposed to be a leaderboard. :) See my issue #1406. I would order them by date. Or you can have the two options with just an arrow next to date & amount so people can pick. But in any case not randomized.

GriffGreen:
Yeah, @krrisis with the CSV feature in place, we can make this a true leaderboard

GriffGreen:
Leaderboard should just have:


Rank | Person | Total Donated | Address
#1 | Santa | $40,000 | 0x839395e20bbB182fa440d08F850E6c7A8f6F0780

Then there should be a plus button that shows the history of their donations and shows:

Date | Amount | Currency | USD Equiv | (Delegated or Direct) | (IF DELEGATED) Delegator

markoprljic:
WIP 
<img width="1218" alt="Image 2020-09-08 at 4 24 59 pm" src="https://user-images.githubusercontent.com/5704285/92489258-e1715480-f1ef-11ea-867e-581459e1309b.png">

markoprljic:
Attached new version, cleaner. 
<img width="1195" alt="Image 2020-09-09 at 7 42 37 am" src="https://user-images.githubusercontent.com/5704285/92559065-2341df80-f270-11ea-9f3e-49f4affc6c6e.png">

aminlatifi:
Donation aggregate service is up now on beta 😃
@mahdine Test it by such requests: https://feathers.beta.giveth.io/aggregateDonations/?id=5b39d45e14cec916d00dab20&$limit=2&$skip=1
Notice result has donations  and user object too, and supports pagination.

aminlatifi:
[...]

PR #1477, opened 2020-10-12
Add leader board

mahdine:
refs #1339

### DIFF
--- a/src/components/LeaderBoard.jsx
+++ b/src/components/LeaderBoard.jsx
@@ -1,0 +1,249 @@
+import React, { Component, Fragment } from 'react';
+import Avatar from 'react-avatar';
+import PropTypes from 'prop-types';
+import moment from 'moment';
+import { Link } from 'react-router-dom';
+
+import config from 'configuration';
+import Loader from './Loader';
+import { getUserName, getUserAvatar, convertEthHelper } from '../lib/helpers';
+import Donation from '../models/Donation';
+
+/**
+ * Shows a table of aggregateDonations for a given type (dac, campaign, milestone)
+ */
+
+class LeaderBoardItem extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      showDetails: false,
+      itemType: null, // Delegated ro Directly donated
+      hasHistory: this.props.d.status === Donation.PAID, // PAID donations always has history
+    };
+
+    this.toggleDetail = this.toggleDetail.bind(this);
+    this.setItemType = this.setItemType.bind(this);
+    this.setItemHasHistory = this.setItemHasHistory.bind(this);
+  }
+
+  setItemType(type) {
+    this.setState({
+      itemType: type,
+    });
+  }
+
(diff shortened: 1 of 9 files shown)


aminlatifi:
Good progress 👍 
image

aminlatifi:
Round this numbers to cent, as you should format it to fit Marko design I suggest Bignumber.js format method
image

aminlatifi:
The branch name prefix should be f_1339_

src/components/LeaderBoard.jsx
```diff
@@ -0,0 +1,249 @@
...
+ */
+
+class LeaderBoardItem extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      showDetails: false,
+      itemType: null, // Delegated ro Directly donated
+      hasHistory: this.props.d.status === Donation.PAID, // PAID donations always has history
```
aminlatifi, line 23: No PAID donation is sent from backend... every where status is compared to not be PAID should be removed

src/components/LeaderBoard.jsx
```diff
@@ -0,0 +1,249 @@
...
+
+  setItemHasHistory(hasHistory) {
+    this.setState({
+      hasHistory,
+    });
+  }
+
+  toggleDetail() {
+    this.setState(prevState => ({
+      ...prevState,
```
aminlatifi, line 45: This line should be deleted

src/components/LeaderBoard.jsx
```diff
@@ -0,0 +1,249 @@
...
+          }
+
+          if (d.status === Donation.PAID || this.state.itemType === 'delegated') {
+            etherScanLink = etherscan && d.txHash ? `${etherscan}tx/${d.txHash}` : '';
+          } else if (this.state.itemType === 'direct') {
+            etherScanLink =
+              homeEtherscan && d.homeTxHash ? `${homeEtherscan}tx/${d.homeTxHash}` : '';
+          }
+          return (
+            <tr key={donation._id} style={this.state.showDetails ? {} : { display: 'none' }}>
```
aminlatifi, line 107: display 'none' is not a good solution on react... you can render the row if showDetails is true.

src/components/LeaderBoard.jsx
```diff
@@ -0,0 +1,249 @@
...
+                <span> {getUserName(giver)}</span>
+              </Link>
+            )}
+          </td>
+          <td className="td-donations-amount font-weight-bold">${totalAmount}</td>
+          <td />
+          <td className="" />
+        </tr>
+        {donations.map(dd => {
+          const donation = new Donation(dd);
```
aminlatifi, line 75: - Is it necessary to create Donation object?
- Please create a separate component for showing donation rows, it's necessary since we need to get donation type (direct/delegate) by using another module and passing it some setter function

aminlatifi (changes requested):

aminlatifi (approved):

PR #1509, opened 2020-11-06
get native currency list from specific whitelist

mdehghani:
closes issue #1418

### DIFF
--- a/src/components/Balances.jsx
+++ b/src/components/Balances.jsx
@@ -36,7 +36,7 @@ class Balances extends Component {
     const { entity } = this.props;
     return (
       <WhiteListConsumer>
-        {({ state: { tokenWhitelist } }) => (
+        {({ state: { nativeCurrencyWhitelist } }) => (
           <div className="dashboard-table-view">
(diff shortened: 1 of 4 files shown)


src/components/Balances.jsx
```diff
@@ -68,8 +68,11 @@ class Balances extends Component {
                           <td className="td-donations-amount">
                             {roundBigNumber(
                               dc.currentBalance / (this.state.conversionRates[dc.symbol] || 1),
-                              (tokenWhitelist.find(t => t.symbol === this.state.currency) || {})
-                                .decimals || 2,
+                              (
+                                nativeCurrencyWhitelist.find(
```
aminlatifi, line 72: Let's change userProvider to solve this once! it's not cool to every time iterate over this list to get the currency precision

aminlatifi (changes requested):
The code works correctly in my tests, just I believe needs improvement.

src/components/Balances.jsx
mdehghani: Done

PR #1523, opened 2020-11-12
H 1413 custom thanks message new pull request 

horus2021:
I have merge conflicts with old h 1413 pull request. I cannot fix it so I create a new pull request as a quick fix.

### DIFF
--- a/src/components/DonateButton.jsx
+++ b/src/components/DonateButton.jsx
@@ -1096,7 +1096,7 @@ export default class Root extends React.PureComponent {
             title: 'Thank you!',
             content: el,
             icon: 'success',
-            buttons: 'No Thanks',
+            buttons: 'OK',
           });
         } else {
           // normal Thanks
--- a/src/components/DonateButton.jsx
+++ b/src/components/DonateButton.jsx
@@ -1079,23 +1079,44 @@ export default class Root extends React.PureComponent {
 
   afterSuccessfulDonate() {
     const { donateToDefaultDac } = this.state;
+    const { customThanksMessage } = this.props.model;
+    // const  customThanksMessage  = 'yes_CS';
 
+    const el = document.createElement('div');
+    el.innerHTML =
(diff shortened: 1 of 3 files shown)


src/components/DonateButton.jsx
```diff
@@ -1082,23 +1082,38 @@ export default class Root extends React.PureComponent {
...
-          title: 'Thank you!',
-          text: 'Would you like to support Giveth as well?',
-          icon: 'success',
-          buttons: ['No Thanks', 'Support Giveth'],
-        }).then(result => {
-          if (result) {
-            this.defaultDacDonateButton.current.openDialog();
-          }
-        });
+        if (typeof customThanksMessage !== 'undefined') {
```
aminlatifi, line 1093: `customThanksMessage` is shown only if we have `donateToDefaultDac`

aminlatifi (changes requested):

src/components/DonateButton.jsx
horus2021: donateToDefualtDac means that user has no profile. If he has no profile, we suggest him to fill his profile. For user without profile who donate to custom message campaign we have the following options.
1. show custom message with more info e.g. Common Stack
2. suggest him to fill his profile (as my commit)
3. show more info e.g Common Stack offer and suggest him to fill his profile

It seems to me  (2) is the best option. (3) suggesting two things is complicated UX. (1) is low priority comparing to (2). It is my opinion, what do you think?

horus2021: - if know user{
-       if custom thanks
-          show custom thanks
-       elsif Donate to Default
-          suggest donate to default 
-        else 
-           show simple thanks
- else //anon user
-     suggest fill profile

aminlatifi: Yes it's the correct scenario

horus2021: @aminlatifi  please check out  my new commit.

Issue #1414, opened 2020-08-13
3% of Donation is below minimum (acceptable) donation


horus2021:
In milestone where 3% of donation goes to another DAC, when 3% donation is too small (0.000000) then we got an error. In the case you donate 0.0000001 ETH in the following milestone you can see this issue. A simple solution can be avoiding 3% to DAC when the value is too small. 

**https://develop.giveth.io/campaigns/5d2eb691e31bd24ef8bd3fa8/milestones/5eaf26784c93f3161c3b73bc**

GriffGreen:
#953 #1000 #1397 #1229

GriffGreen:
If the value is small the 3% is ignored

GriffGreen:
When the user clicks donate we should have a check that skips the donation to the DAC if it's too small... cut off donation should be around $1... each token has it's own precision so that should be considered

GriffGreen:
To be handled in the front end

GriffGreen:
$40 should be the limit here... this will give us a $1.20 donation

GriffGreen:
We need to put a limit on the lower end.... this should be fast

mohammadranjbarz:
@MoeNick  for testing this you should create a trace and mark the `Donate 3% to Giveth`
then for donations with amount 10$ or less than 10$ your all donations will go to trace, 
but for greater amount the 3% percent of donations should go to community

PS 1 : the amount in production is 50$ ( it doesn't matter what token you donate, we compare the usdValue)
PS 2: this 3 percent works for donation not delegation

MoeNick:
It worked for me. but for second scenario (less that 10) there is a UX problem, the donation modal still be there for minutes before the submitting trx showed up. 
Let me test it more it may lead to another UI problem over there.

PR #1539, opened 2020-11-23
gh-1414 min amount for delegating to DACs

arbreton:


src/components/DonateButton.jsx
```diff
@@ -184,7 +184,10 @@ class DonateButton extends React.Component {
 
     let { maxDonationAmount } = this.props;
     if (maxDonationAmount) {
-      if (dacId !== undefined && dacId !== 0) {
+      const amountDAC = parseFloat(maxDonationAmount - maxDonationAmount / 1.03)
```
aminlatifi, line 187: If you want to get the amount will be donated to dac, can simply write `maxDonationAmount * 0.03`

aminlatifi (changes requested):
The implementation doesn't address issue #1414 requirement. You should obtain the usd value of original donation and compare it with 40$. If it's less than that ignore donating to DAC

arbreton:
Will check on the USD value

aminlatifi:
We are implementing an api on feathers for hourly rates, you can use it after is merged
https://github.com/Giveth/feathers-giveth/pull/287

arbreton:
Sounds great, will do

PR #1544, opened 2020-11-26
F 771 link preview

aminlatifi:


aminlatifi:
The link preview service will be implemented by different design. So the changes in this PR will not be merged.

PR #1568, opened 2020-12-03
Dont show delegate button when there is no options to select

### LINKED ISSUE #1536: Delegate Funds from a DAC Option on viewing a Campaign
Arriving at a Campaign, I have the option to Delegate funds from a DAC. 
When I click on it, it says there are no options - in which case the button should not appear as the user does not have a DAC from which to Delegate to the Campaign. Maybe because this account is whitelisted to be able to make DACs? 
The Window does not have a cancel button, so I felt like I got stuck there..  which I was, you can't click away from the popup box to make it go away. Escape key did let me escape :)

image

mohammadranjbarz:
fix #1536

Don't show delegate button when there is no options to select, instead of the de button return some text,
I think it's good but if you have another idea , I'm all ears.

image


src/components/views/ViewMilestone.jsx
```diff
@@ -160,6 +163,40 @@ class ViewMilestone extends Component {
     );
   }
 
+  async shouldShowDelegateButton() {
+    const { currentUser } = this.props;
+    const { campaign } = this.state;
+    if (!campaign || !currentUser) {
```
mohammadranjbarz, line 169: Some data is not loaded when the component did mount, so I cant calculate the permissions when component created, and for that particular reason we should check if all data doesn't provided we don't proceed, when the data is provided then we check the permissions

src/services/WhitelistService.js
```diff
@@ -0,0 +1,15 @@
+import { feathersClient } from '../lib/feathersClient';
```
aminlatifi, line 1: Don't add it, we have whitelist provider already... use that.

aminlatifi (changes requested):

src/services/WhitelistService.js
mohammadranjbarz: If you mean this

https://github.com/Giveth/giveth-dapp/blob/e8b4938644a07e99d8b0ed1063a41d9d8c29e3fc/src/contextProviders/WhiteListProvider.jsx#L19-L30


This function is not exported so I cant use that.

aminlatifi: WhitelistProvider export few functions, one is `isDelegate` which returns whether the user is a delegate or not

mohammadranjbarz:
I removed the Whitelist service and instead I used WhitelistProvider , but I think my cods need some reviews, although I tested and it was ok.

src/containers/Application.jsx
```diff
@@ -156,252 +156,257 @@ class Application extends Component {
                                         {!userLoading && !hasError && (
                                           <div>
                                             <MainMenu />
-
-                                            <Switch>
-                                              {/* Routes are defined here. Persistent data is set as props on components
+                                            <WhiteListConsumer>
```
aminlatifi, line 159: No, remove it.

src/components/views/ViewCampaign.jsx
```diff
@@ -518,6 +548,7 @@ ViewCampaign.propTypes = {
     }).isRequired,
   }).isRequired,
   balance: PropTypes.instanceOf(BigNumber),
+  isDelegate: PropTypes.func.isRequired,
```
aminlatifi, line 551: Don't pass isDelegate as a property... use new useContext hook.

aminlatifi: Or use consumer immediately above the component which needs whitelist provider values

mohammadranjbarz:
All changes of `Application.jsx` reverted.

aminlatifi:
@mohammadranjbarz Thanks for implementing... but it seems too complicated.
How about moving all the process of user is delegate into user provider? I mean user entity provided by user provider hold some property indicating whether user is delegate or not. This will reduce code size and make it simpler

mohammadranjbarz:
> @mohammadranjbarz Thanks for implementing... but it seems too complicated.
> How about moving all the process of user is delegate into user provider? I mean user entity provided by user provider hold some property indicating whether user is delegate or not. This will reduce code size and make it simpler

@aminlatifi  You are right it's complicated and have some repetitive code, 
I'll work on it.

mohammadranjbarz:
> @mohammadranjbarz Thanks for implementing... but it seems too complicated.
> How about moving all the process of user is delegate into user provider? I mean user entity provided by user provider hold some property indicating whether user is delegate or not. This will reduce code size and make it simpler


If I understood your opinion you are meaning we add isDelegate function to User provider so  we can get the `isDelegate` from actions in here:
https://github.com/Giveth/giveth-dapp/blob/1c3ced0260680066e839e11fb9ef27610d04d659/src/containers/Application.jsx#L148

and then we can pass it as props to `ViewMilestone` and `ViewCampaign` then we don't need adding `WhiteListProvider` to `ViewCampaign` and `ViewMilestone`

mohammadranjbarz:
@aminlatifi 
Your requested changes has been applied now isDelegate function is in UserProvider so we have less changes in jsx files, and now its more easy and safe to merge

src/components/views/ViewMilestone.jsx
```diff
@@ -160,6 +162,34 @@ class ViewMilestone extends Component {
...
+    // aren't undefined proceed checking
+    this.setState({ checkShowDelegation: true });
+    const userAddress = currentUser && currentUser._address;
+    if (campaign.ownerAddress === userAddress) {
+      this.setState({ showDelegateButton: true });
+      return;
+    }
+
+    // should check if user has any dac
+    const userDacs = (await DACService.getDACsOwnedByUser(userAddress)).data;
```
aminlatifi, line 181: I imagined moving all this logic to user provider too.

mohammadranjbarz: @aminlatifi  Ooops you are right, it can be moved too.

mohammadranjbarz: @aminlatifi 
Now all logic moved to user provider.

src/components/ViewMilestoneAlerts.jsx
```diff
@@ -12,81 +12,88 @@ import ChangeMilestoneRecipientButton from './ChangeMilestoneRecipientButton';
...
-        <ProjectViewActionAlert message="Delegate some donation to this project">
-          <DelegateMultipleButton
-            milestone={milestone}
-            campaign={campaign}
-            balance={balance}
-            currentUser={currentUser}
-          />
-        </ProjectViewActionAlert>
-      )}
+    <UserProvider account={currentUser && currentUser.address}>
```
aminlatifi, line 23: It's wrong way of using provider

aminlatifi (changes requested):
Please fix conflicts with latest develop version too (By rebase or merge)

src/components/ViewMilestoneAlerts.jsx
mohammadranjbarz: @aminlatifi Could you link me a good way in our project to I modify this like that?

aminlatifi: https://github.com/Giveth/giveth-dapp/blob/c5d053b1303750f1fca79ed42d141929341ca84d/src/components/views/Milestones.jsx#L17

aminlatifi: Above is how the value of UserProvider is consumed.

GriffGreen:
Merge conflicts.... it needs to be redone on a new branch more manually, @aminlatifi is working on this today

Issue #1532, opened 2020-11-17
UI Text Revision: Create A Campaign


Danibelle:
Text instructions on this page are not clear and lead to very long DAC names.  In part, because, as soon as you click away from the Title text box the line of text instructions below it disappears!
![image](https://user-images.githubusercontent.com/39466970/99337033-28f70b00-2847-11eb-9ad9-1508c742455d.png)
![image](https://user-images.githubusercontent.com/39466970/99338249-74aab400-2849-11eb-9ca2-2944a68ecf10.png)

**Current Header Text:** A DAC aims to solve a cause by building a community, raising funds, and delegating those funds to Campaigns that solve its cause. Should you create a Campaign or Community? Read more here.
**Recommended Text Revision:** A DAC unites Givers and Makers in building a community around their common vision to raise then delegate funds to Campaigns that deliver a positive impact to shared goals.

**Current DAC Title Box Text:** Community cause *

**Recommended Text Revision:** Name your Community *
Describe your altruistic community (DAC) in 1 sentence
^^make it so the instruction text remains even if the box gets highlighted RED as it does^^

**Current DAC Description Box Text:** Explain your cause *
Make it as extensive as necessary. Your goal is to build trust so that people join your Community and/or donate Ether.
**Recommended Text Revision:** Explain the cause of your community *
Describe the shared vision and goals of your Community and the cause that you are collaborating to solve. Share links, insert media to convey your message and build trust so that people will join your Community and/or donate to the cause.

horus2021:
@aminlatifi please checkout my [PR](https://github.com/Giveth/giveth-dapp/pull/1587)
@Danibelle  You are great copywriter! Thanks.

aminlatifi:
@horus2021 Did you test on develop? beta is not merged yet!

Danibelle:
Tested, changes confirmed - thank you!

PR #1587, opened 2020-12-08
Issue 1532 Update Text for Create DAC form

horus2021:
#1532 Revise Text for Create DAC form

### DIFF
--- a/src/components/views/EditDAC.jsx
+++ b/src/components/views/EditDAC.jsx
@@ -262,7 +262,7 @@ class EditDAC extends Component {
                       label="Name your Community"
                       type="text"
                       value={dac.title}
-                      placeholder="Describe your altruistic community (DAC) in 1 sentence"
+                      placeholder="e.g. Hurricane relief."
                       help="Describe your Decentralized Altruistic Community (DAC) in 1 sentence."
                       validations="minLength:3"
                       validationErrors={{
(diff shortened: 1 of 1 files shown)


src/components/views/EditDAC.jsx
```diff
@@ -259,10 +259,10 @@ class EditDAC extends Component {
                     <Input
                       name="title"
                       id="title-input"
-                      label="Community cause"
+                      label="Name your Community"
                       type="text"
                       value={dac.title}
-                      placeholder="e.g. Hurricane relief."
```
aminlatifi, line 265: This should not be changed

src/components/views/EditDAC.jsx
```diff
@@ -259,10 +259,10 @@ class EditDAC extends Component {
...
+                      placeholder="Describe your altruistic community (DAC) in 1 sentence"
                       help="Describe your Decentralized Altruistic Community (DAC) in 1 sentence."
```
aminlatifi, line 266: change attribute to `helpText` and set the value "Describe your altruistic community (DAC) in 1 sentence"

aminlatifi, line 266: It's the one which should be changed

horus2021: @aminlatifi  You are right see new commit.

Issue #730, opened 2019-03-24
Get rid of the waiting for donations in tokens


GriffGreen:
It would be pretty easy to reduce the pain of donating tokens by allowing the user to submit both txs at once. 

I would suggest we just have a session where we revise the donation design, allowing the user to 

1. unlock (better term than approve) just the amount that they need for this donation (default)

2. Flip a switch that says something along the lines of "unlock 10x the amount of this donation, this is a risk but because i trust the Giveth team and Giveth smart contracts i would rather enjoy a better UX dont want to have to pay extra gas everytime i want to donate"  this will multiply the donation by 100 and approve their account for that amount.... it is important that if there is an approval for an amount larger than the

3. GET RID OF WAITING! Allow the user to click a pay button to make the payment tx right away so they can get on with their day.

GriffGreen:
follows #735

GriffGreen:
We need to hot wire metamask to auto estimate gas for the second tx and have it pop up right away

GriffGreen:
set the gas limit to 400,000 for the second gaslimit

aminlatifi:
https://docs.metamask.io/guide/sending-transactions.html#example

GriffGreen:
tackling this this week

arbreton:
some updates, the transactions are working OK now, seem to be a problem with token amount specifics that need some more investigation, will reach out tomorrow morning

GriffGreen:
3% milestone is worth testing

aminlatifi:
Any updates?

arbreton:
Still looking at this one for some reason 2nd tx is failing even when gas is very high and tokens are approved could be a bug

aminlatifi:
Make gas limit 200K

aminlatifi:
@GriffGreen I tried to do the donate tx immediately after approve allowance, but triggering the tx cause this exception from MetaMask and no popup is shown
```
https://home.etherscan.io/tx/undefined => Error: Internal JSON-RPC error.
{
  "message": "VM Exception while processing transaction: revert",
  "code": -32000,
  "data": {
    "stack": "c: VM Exception while processing transaction: revert\
    at Function.c.fromResults (/Users/amin/Work/giveth/feathers-giveth/node_modules/ganache-cli/build/ganache-core.node.cli.js:4:194812)\
[...]

PR #1595, opened 2020-12-14
gh-730 get rid of the waiting on donations

arbreton:


src/components/DonateButton.jsx
```diff
@@ -482,31 +483,24 @@ class DonateButton extends React.Component {
...
-          );
+        this.donateWithBridge(
+          dacId,
+          amountDAC,
+          donationOwnerAddress,
+          amount,
+          comment,
+          allowanceApprovalType,
+        );
+        setTimeout(() => {
```
aminlatifi, line 496: This assumes user will click confirm on his wallet in in one second after first tx popup... so app will show the second popup 1 second after the first one...
It's not good way of implementing it...
The eligible solution is to showing the second popup after user has confirmed the first one.

src/components/DonateButton.jsx
```diff
@@ -671,19 +675,24 @@ class DonateButton extends React.Component {
...
         }
-        const allowed = await DonationService.approveERC20tokenTransfer(
+
+        const allowed = DonationService.approveERC20tokenTransfer(
           tokenAddress,
           currentUser && currentUser.address,
           allowanceRequired.toString(),
         );
 
+        setTimeout(() => {
```
aminlatifi, line 685: The same as above

aminlatifi (changes requested):

Issue #1524, opened 2020-11-12
We need a progress bar for uploading images


GriffGreen:
it takes time to upload images... we need to tell the user

PR #1621, opened 2020-12-31
add loading on uploading images quillForm #1524.

peymannaderi3:


aminlatifi (changes requested):
Please add these features:
1. Please resize images to become smaller size (we need this feature somewhere else)
2. Consider image removing from conversation during upload time

peymannaderi3:
Add a loading overlay on text editor when uploading images.

Issue #1464, opened 2020-09-30
Sort by activity and Hide inactive Campaigns and DACs from the front page


GriffGreen:
Currently, we are featuring campaigns and DACs that haven't had any activity in years, the sorting on the front page should be by activity, and projects that haven't had activity for 6+mo should be hidden off the front page...
<img width="1269" alt="Screen Shot 2020-09-30 at 4 32 09 PM" src="https://user-images.githubusercontent.com/14945613/94699344-83d9af00-033a-11eb-823a-029a9ec84a2c.png">

arbreton:
After some research on the code got the logic for sorting on the main page, now just querying the database for 6 months newer project should do the job, diving into MongoDB docs for date ranges

PR #1622, opened 2020-12-31
hide activity for 6+mo #1464.

peymannaderi3:


src/models/BasicModel.js
```diff
@@ -40,6 +40,7 @@ class BasicModel extends Model {
     donationCounters = [],
     token,
     createdAt,
+    updatedAt,
```
aminlatifi, line 43: Is it necessary to add this?

aminlatifi: Please test without adding updatedAt here and other places. Just keep it in query.

peymannaderi3: It's not necessary. I added it to check the list dates.

PR #1653, opened 2021-01-10
H 1625 metamask errors

peymannaderi3:
Add a wrapper function for metamask to only log the errors instead of showing the popup.


aminlatifi:
Will be implemented by another approach and different PRs

Issue #1666, opened 2021-01-12
Fix interactivity problems of user profile form


aminlatifi:
User profile form has several problems like
- [x] Save button is enabled when form is not valid
- [x] Sometimes the tx not happens and maybe not save to DB, but user sees successful update notification
- [ ] User native currency doesn't need to be saved to network, so if it's the only change we don't need to push it to network.
- [x] Don't save user native currency to smart contract and IPFS, and if the modification is only that field used doesn't need to do a transaction or uploading IPFS

PR #1673, opened 2021-01-14
H 1666 fix user profile form

peymannaderi3:


peymannaderi3:
#1666

aminlatifi:
@Thanks, it as we expected
There is a bug which appears in this scenario:

1. Updated Email
2. Pressed submit button
3. Did the transaction
4. New data is uploaded as new object to IPFS and the hash value is saved in smart contract
5. But the database (feathers) `url` entry (IPFS value) is not updated.

aminlatifi:
> @thanks, it as we expected
> There is a bug which appears in this scenario:
> 
> 1. Updated Email
> 2. Pressed submit button
> 3. Did the transaction
> 4. New data was uploaded as new object to IPFS and the hash value was saved in smart contract
> 5. But the **database (feathers) `url` entry (IPFS value) was not updated.**

I was wrong, URL field is updated later in backend.

aminlatifi:
Seems OK

PR #1791, opened 2021-02-12
add slug to DAC

mdehghani:


mohammadranjbarz:
@mdehghani 
I think it would be better to show more clear message when user write a repetive text for milestone title

image

mohammadranjbarz:
image

when proposing new milestone we have id campaignId in url, is it possible to have slug over there?

mohammadranjbarz (changes requested):
It it not supposed to support older links like:
`/campaigns/campaignId/milestones/milestoneId`?
Beucase I think all links that sent to our users is something like that, and if we want deprecate those urls we should change URLs that we send in emails

mdehghani:
> @mdehghani
> I think it would be better to show more clear message when user write a repetive text for milestone title
> 
> image

It would be fixed in another PR

mdehghani:
> image
> 
> when proposing new milestone we have id campaignId in url, is it possible to have slug over there?

As discussed to Amin, we decided to use slug only in 'view' urls.
What do you think Amin? @aminlatifi

mdehghani:
> It it not supposed to support older links like:
> `/campaigns/campaignId/milestones/milestoneId`?
> Beucase I think all links that sent to our users is something like that, and if we want deprecate those urls we should change URLs that we send in emails

Yes, we are supporting old urls now. The mentioned url is supported and will be redirected to the new one. Is there a problem with you?

aminlatifi:
> It it not supposed to support older links like:
> `/campaigns/campaignId/milestones/milestoneId`?
> Beucase I think all links that sent to our users is something like that, and if we want deprecate those urls we should change URLs that we send in emails

In another PR we should redirect all previous links to new one (with slug)

aminlatifi:
> > image
> > when proposing new milestone we have id campaignId in url, is it possible to have slug over there?
> 
> As discussed to Amin, we decided to use slug only in 'view' urls.
> What do you think Amin? @aminlatifi

Yes, it's ok so far. Let's keep it

Issue #1786, opened 2021-02-10
User sees /profile page without unlocking wallet (MetaMask)


Santigs:
The app does not seem to react when I click the send button

aminlatifi:
@Santigs Would you please provide more information? Which send button?

Santigs:
Sorry, I meant the Save button
![Screenshot 2021-02-10 at 21 38 32](https://user-images.githubusercontent.com/34634436/107612801-576a6380-6c47-11eb-9496-58242c2c092c.png)
The page does not seem to react. I provided info in all fields

aminlatifi:
@Santigs It seems you haven't enabled Web3, basically you shouldn't see this page until you connect your MetaMask. Thanks for reporting this.

aminlatifi:
@Santigs To make it go through you should unlock your metamask and then try again

GriffGreen:
If the wallet is locked, and the user clicks the Save Profile button, they should be asked to log into metamask.

aminlatifi:
I prefer something like this
https://github.com/MohammadPCh/giveth-dapp/blob/f_1730_create_new_UI/src/components/Web3ConnectWarning.jsx
@MohammadPCh I guess it's better to keep consistency, so fix this issue as part of above branch

WhyldWanderer:
No matter how hard I try,  i am unable to even create a profile without enabling web3 so i assume the loophole has been closed. (Ive tried through creating a new campaign or milestone or donating)

I get this message
![image.png](https://images.zenhubusercontent.com/5ffdde8079760c4346381d45/21bf814a-be67-4110-9263-1e3903fcef71)

When this message pops up, there is no [x] close window option so i must press the back button in my browser to remove the pop up.

MoeNick:
@WhyldWanderer it's not have x button because you need to use metamask changing your account I assume. What do yo think?

aminlatifi:
> No matter how hard I try, i am unable to even create a profile without enabling web3 so i assume the loophole has been closed. (Ive tried through creating a new campaign or milestone or donating)
> 
> I get this message
> ![image.png](https://camo.githubusercontent.com/3809456a801efec5f292e9960fd96b1de7835ddd6b69f24246a6e1d5a4893cab/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3566666464653830373937363063343334363338316434352f32316266383134612d626536372d343131302d393236332d316533393033666365663731)
> 
[...]

PR #1827, opened 2021-02-26
F 1786 ask user to unlock wallet to save profile

MohammadPCh:
refs #1786 
@aminlatifi can you help me about UX writing in this issue? because i'm not familiar with terminology of Giveth.
Thanks.


aminlatifi:
I prefer something like this
https://github.com/MohammadPCh/giveth-dapp/blob/f_1730_create_new_UI/src/components/Web3ConnectWarning.jsx
@MohammadPCh I guess it's better to keep consistency, so fix this issue as part of above branch

Issue #1837, opened 2021-03-03
Not showing status updates when it's delegated from DAC


mohammadranjbarz:
https://beta.giveth.io/campaigns/5fbbce04188f980d941470eb/milestones/601abd93f9e6e877930e9694

![image](https://user-images.githubusercontent.com/9850545/109862373-9c494f00-7c75-11eb-8635-c6057562ca8f.png)

aminlatifi:
It needs a fix in backend

PR #1838, opened 2021-03-03
Fix showing status update for DAC delgates

mohammadranjbarz:
related to #1837

### DIFF
--- a/src/components/MilestoneConversationItem.jsx
+++ b/src/components/MilestoneConversationItem.jsx
@@ -85,7 +85,7 @@ const getReadableMessageContext = conversation => {
           </React.Fragment>
         );
       }
-      if (donorType === 'dac') {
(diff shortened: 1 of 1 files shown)


src/components/MilestoneConversationItem.jsx
```diff
@@ -85,7 +85,7 @@ const getReadableMessageContext = conversation => {
           </React.Fragment>
         );
       }
-      if (donorType === 'dac') {
+      if (donorType === 'giver') {
```
aminlatifi, line 88: It cannot be the solution, we need to fix it in backend to send DAC info. By this change DAC info and User role is not shown in conversation.

mohammadranjbarz: @aminlatifi 
As we talked I reopened this PR and I made some changes

Issue #1882, opened 2021-03-11
Misspelling InProgress instead of In Progress in Details Box of Milestone Page


horus2021:
![InProgress --- In Progress](https://user-images.githubusercontent.com/65175075/110794346-e7252100-828a-11eb-9fcf-9ad484546fc8.JPG)

96RadhikaJadhav:
@horus2021 Can I work on the issue?

aminlatifi:
> @horus2021 Can I work on the issue?

Hi @96RadhikaJadhav 

Thanks for your intention to contribute to our Dapp 🙏  , you can work on this issue 😃

96RadhikaJadhav:
@aminlatifi @horus2021 I have created PR #1906 . Could you please check and let me know if any changes required.

aminlatifi:
Thanks @96RadhikaJadhav, you are very kind since decided to contribute to our DApp

PR #1906, opened 2021-03-14
Corrected InProgress typo

96RadhikaJadhav:
This PR will resolve issue #1882 .


aminlatifi:
@96RadhikaJadhav InProgress is status value which is saved in backend, so it cannot just simply changed in the way you did

Issue #1917, opened 2021-03-15
Update milestone donations list to represent final payment info


aminlatifi:

aminlatifi:
The status link should point to Home Tx when is paid really
The status title should be:
- In Processing
- Paid

aminlatifi:
Tested and works.

PR #1944, opened 2021-03-18
Update milestone donations list to represent final payment info

mohammadranjbarz:
related to #1917

I suggest to run it locally with this command :` REACT_APP_ENVIRONMENT=beta  npm start` because  in beta you can see the statuses better.

### DIFF
--- a/src/models/Donation.js
+++ b/src/models/Donation.js
@@ -104,6 +104,8 @@ class Donation extends Model {
     this._usdValue = data.usdValue;
     this._parentDonations = data.parentDonations;
     this._comment = data.comment;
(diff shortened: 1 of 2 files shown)


src/models/Donation.js
```diff
@@ -165,7 +167,11 @@ class Donation extends Model {
       case Donation.PAYING:
         return 'paying';
       case Donation.PAID:
-        return 'paid';
+        if (this._bridgeStatus === Donation.PAID) {
```
mohammadranjbarz, line 170: Maybe if bridgeStatus is Expired we should return `paid` as well

Issue #1930, opened 2021-03-16
Add banner on front page explaining transaction limit


laurenluz:
Spoke with @Danibelle about this. Considering issues #1885 #1886 #1887 we think it makes sense to give the user a heads-up about the limitation on transaction values... so they're not just finding out last-minute when they try to make, disburse, collect from, or mark complete a milestone.

Suggestion to add a banner to the front page that says:
"Giveth pays the gas fees so you don't have to! Because of this we are temporarily limiting transactions to those greater than ${value} ${token.symbol}"

@markoprljic - what do you think? can we add something like this?

markoprljic:
UX approved. 😊

laurenluz:
ballin'. @aminlatifi - can you help make this happen?

aminlatifi:
@laurenluz It's not clear to me `add a banner to the front page`

laurenluz:
hey @aminlatifi something simple like this "free delivery" banner at the top of this clothing page... any maybe with a little x in the corner so the user can hide the banner after reading.

@Danibelle - what do you think? is this what you had in mind?
@markoprljic is there a specific colour that you think we could use? maybe a simple design would be helpful.

![image](https://user-images.githubusercontent.com/75490971/111919172-35d27800-8a4e-11eb-9632-e0e284de627b.png)

Danibelle:
Exactemente!

GriffGreen:
Anyone who comes to Giveth Trace, they should know that there is a big change here.

markoprljic:
Here's the banner, let me know what you think. @Danibelle @laurenluz @aminlatifi @GriffGreen 

![Image 2021-03-22 at 6 40 50 pm](https://user-images.githubusercontent.com/5704285/112033917-334f4b80-8b3e-11eb-8a78-9e41f88f94ee.jpg)

laurenluz:
I think it's awesome 😍

Danibelle:
Danibelle Approved!!
(Can I get a banner for that? 🤔😉🤣)

markoprljic:
It's here for devs to grab and implement. https://www.figma.com/file/SuQbqtNjtRNrlYGAXijaMI/Giveth-TRACE?node-id=2072%3A827

MoeNick:
@RamRamez the text is ok.. we got value form backend and we just put USD as TokenSymbol.

MoeNick:
[...]

PR #1966, opened 2021-04-01
F 1930 add banner explaining transaction limit

RamRamez:
Related to issue #1930


MoeNick:
Banner showed up at the bottom instead of top
image

RamRamez:
> Banner showed up at the bottom instead of top
> image

Putting banner on top would add unnecessary complexity to the UI since current menu component don't let us much customizations. So we are going to change the Menu bar design to Ant Design. But for now we can merge the pull request to inform users about the current tx limitations, or if we need banner on top, we can keep this issue open till completing new menu design.

Issue #1975, opened 2021-04-04
New main menu with Ant Design


RamRamez:
New main menu with Ant Design

PR #1978, opened 2021-04-05
F 1975 ant design main menu

RamRamez:
Related to Issues #1975 and #1930 
It's the initial version for testing.


aminlatifi:
Report Issue is not center
Mobile menu is not shown on mobile view

aminlatifi:
@RamRamez 

Report Issue button is not center!


image

src/components/layout/MainMenu/LeftMenu.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+  let selectedKey;
+  if (isDac) selectedKey = 'Communities';
+  else if (isCamp) selectedKey = 'Campaigns';
+  else if (isMilestone) selectedKey = 'Milestones';
+  else selectedKey = '';
+
+  useEffect(() => {
+    setSelectedKeys(selectedKey);
+  }, [selectedKey]);
```
aminlatifi, line 23: Make this useEffect dependent to path. Maybe location parameter passed by `withRouter`
```suggestion
  }, [props.location]);
```

> A location object is never mutated so you can use it in the lifecycle hooks to determine when navigation happens, this is really useful for data fetching and animation.

If above works, all computation of other variables should be happen in useEffect, I mean selectedKey, isDac and etc.

>

src/components/layout/MainMenu/LeftMenu.jsx
```diff
@@ -0,0 +1,40 @@
...
+        <Link to="/campaigns">Campaigns</Link>
+      </Menu.Item>
+      <Menu.Item key="Milestones">
+        <Link to="/milestones">Milestones</Link>
+      </Menu.Item>
+    </Menu>
+  );
+};
+
+export default LeftMenu;
```
aminlatifi, line 40: Use React memo to avoid unnecessary re-rendering this component. Maybe isEqual function should be defined with comparing props.location

src/components/layout/MainMenu/MainMenu.jsx
```diff
@@ -0,0 +1,69 @@
...
+            <LeftMenu />
+            <RightMenu />
+          </Drawer>
+        </div>
+      </div>
+    </nav>
+  );
+};
+
+export default MainMenu;
```
aminlatifi, line 69: Use react memo here

src/components/layout/MainMenu/RightMenu.jsx
```diff
@@ -0,0 +1,136 @@
...
+  else if (myMilestones) selectedKey = 'Manage:1';
+  else if (donations) selectedKey = 'Manage:2';
+  else if (delegations) selectedKey = 'Manage:3';
+  else if (myDacs) selectedKey = 'Manage4';
+  else if (myCampaigns) selectedKey = 'Manage:5';
+  else selectedKey = '';
+
+  useEffect(() => {
+    setSelectedKeys(selectedKey);
+  }, [selectedKey]);
```
aminlatifi, line 57: As above, make it depend on props.location

aminlatifi (changes requested):
Great job @RamRamez. Please just do few amendments.

src/components/layout/MainMenu/LeftMenu.jsx
RamRamez: I tried different solutions like useRouteMatch, useLocation and useParams but none of them can be defined inside useEffect because hooks must be defined inside body of functional components.

src/components/layout/MainMenu/MainMenu.jsx
RamRamez: I tried React.memo for MainMenu, it prevents left menu from unnecessary rerendering but not useful for right menu because it uses different contexts for user management menus.
So we're using memo just for a very little component (left menu) that seems inefficient. As a result in this dilemma, I think we can accept some unnecessary rerendering of the left menu instead of memo equality check calculation and browser cache.

PR #1992, opened 2021-04-07
Set limit for every currency for withdraw

mohammadranjbarz:
related to #1887 #1886

PR #2003, opened 2021-04-08
Remove MilestoneConversationItem component

mohammadranjbarz:
related to #1992 
https://github.com/Giveth/giveth-dapp/pull/1992#issuecomment-815575635

I removed that component instead of calculating adding `isAmountEnoughForWithdraw`, because  it was legacy and useless
it's handles the conversations with statuses :
`proposed`, `rejected`, `NeedsReview`, `Completed `, `proposedRejected`
and such events are not in our DB (there are few documents for 2018 and nothing more):
https://feathers.beta.giveth.io/conversations?messageContext=proposed
https://feathers.beta.giveth.io/conversations?messageContext=rejected
https://feathers.beta.giveth.io/conversations?messageContext=NeedsReview
https://feathers.beta.giveth.io/conversations?messageContext=Completed
https://feathers.beta.giveth.io/conversations?messageContext=proposedRejected


aminlatifi:
@mohammadranjbarz There aren't because no one has used, but it's possible. If it's cumbersome and redundancy to compute that value, what about creating a provider/consumer in order to set milestone and other things once and use it everywhere inside milestone view and its children? I am also thinking about subscribe! to put a button in conversation menu!

mohammadranjbarz:
> @mohammadranjbarz There aren't because no one has used, but it's possible. If it's cumbersome and redundancy to compute that value, what about creating a provider/consumer in order to set milestone and other things once and use it everywhere inside milestone view and its children? I am also thinking about subscribe! to put a button in conversation menu!

@aminlatifi 
No I can calculate that, 
but I was thinking changing a component that never used, doesn't make sense, and deleting that component is better.
but if you think maybe we use this sometime we can not merge this PR, 
BTY is this where you said we didnt pass parameter or there are something else also, because I couldn't see logs
https://github.com/Giveth/giveth-dapp/pull/1992#issuecomment-815976033

Issue #1887, opened 2021-03-11
Don't allow mark complete a milestone with less than limit balance


aminlatifi:

laurenluz:
When user clicks "Mark Complete" on this page...

![image](https://user-images.githubusercontent.com/75490971/111365532-376bfc80-8658-11eb-90dd-23c6a265f184.png)

Please check that the donation balance is greater than the limit "${value} ${token.symbol}" (fetched from the backend), and if it is not, have a popup that says:

“Oh No!
A minimum donation balance of ${value} ${token.symbol} is required before you can mark this milestone is complete. This is a temporary limitation due to Ethereum Mainnet issues.”

@mohammadranjbarz can you help to implement this change?

mohammadranjbarz:
> When user clicks "Mark Complete" on this page...
> 
> ![image](https://user-images.githubusercontent.com/75490971/111365532-376bfc80-8658-11eb-90dd-23c6a265f184.png)
> 
> Please check that the donation balance is greater than the limit "${value} ${token.symbol}" (fetched from the backend), and if it is not, have a popup that says:
> 
> “Oh No!
> A minimum donation balance of ${value} ${token.symbol} is required before you can mark this milestone is complete. This is a temporary limitation due to Ethereum Mainnet issues.”
> 
> @mohammadranjbarz can you help to implement this change?

@laurenluz 
Sure Lauren I put it in my tasks

MoeNick:
@RamRamez has reviewed.

MoeNick:
Everey donate-currancy shouldnt be less than 20

Issue #1886, opened 2021-03-11
Don't allow user collect/disburse milestone with less than limit


aminlatifi:

laurenluz:
This is similar to #1887 

When user hits "collect" or "disperse"...

Please check that the donation balance is greater than the limit "${value} ${token.symbol}" (fetched from the backend), and if it is not, have a popup that says:

“Oh No!
A minimum donation balance of ${value} ${token.symbol} is required before you can collect or disperse the funds. This is a temporary limitation due to Ethereum Mainnet issues.”

@mohammadranjbarz passing off to you - can you implement?

mohammadranjbarz:
> This is similar to #1887
> 
> When user hits "collect" or "disperse"...
> 
> Please check that the donation balance is greater than the limit "${value} ${token.symbol}" (fetched from the backend), and if it is not, have a popup that says:
> 
> “Oh No!
> A minimum donation balance of ${value} ${token.symbol} is required before you can collect or disperse the funds. This is a temporary limitation due to Ethereum Mainnet issues.”
> 
> @mohammadranjbarz passing off to you - can you implement?

@laurenluz 
Thanks Lauren I would do it but I thinks this epic is not priority for this sprint, maybe next sprint ( Or anytime @aminlatifi  that Amin says)

MoeNick:
Reminder: baner is still at the bottom of the page

MoeNick:
the error is small and go away quickly. 
![image](https://user-images.githubusercontent.com/80677512/114405173-a850f680-9bbb-11eb-8054-6f85c2e2ef93.png)


Shouldn't it be like the one we have on create milestone? As a user I prefer it in the middle, and needs my confirm "OK" before going away. 
what do u think Amin.

PR #2007, opened 2021-04-11
Pass isAmountEnoghForWithdraw to all sub compoenents

mohammadranjbarz:
related to #1887, #1886

I suggest to instead of this PR you merge the #2003  PR


aminlatifi:
Thanks @mohammadranjbarz for implementing this. You have spent time to implement this and it's valuable to keep the feature, since it has been implemented before and maybe based on some requirements we don't know about it. So keeping this component and meanwhile supporting new feature is good for us.

aminlatifi:
@MoeNick This feature is ready to be tested now

Issue #2011, opened 2021-04-11
Changing donate and delegate amount sliders to Ant Design


RamRamez:
Changing donate and delegate amount sliders to Ant Design

PR #2012, opened 2021-04-12
F 2011 Change amount sliders to Ant Design

RamRamez:
Issue #2011


aminlatifi:
@RamRamez Can you change its thickness? Also use it in delegate modals.

RamRamez:
> @RamRamez Can you change its thickness? Also use it in delegate modals.

Done

Issue #1803, opened 2021-02-17
When delegating from the Commons Stack Association I keep getting this weird notification... but it's not true!


GriffGreen:
<img width="543" alt="Screen Shot 2021-02-16 at 8 13 52 PM" src="https://user-images.githubusercontent.com/14945613/108142336-df2bf400-70c5-11eb-86c0-5f7c645db098.png">

aminlatifi:
The command is generally true, but not applicable for this milestone because it's capped and doesn't need to donate from more than 10 sources.

GriffGreen:
the logic is complicated... but it should be aware that I am not donating more than the first 10 delegations amounts...

aminlatifi:
@RamRamez On develop I see this while is incorrect
![image](https://user-images.githubusercontent.com/5684607/113487795-4b748280-94cf-11eb-8e07-df4d501c65f1.png)

aminlatifi:
I will test it

MoeNick:
@aminlatifi I couldn't reproduce it :(

PR #2036, opened 2021-04-15
H 1803 delegation limit notif

RamRamez:
Issue #1803
1. fix delegation limit notification shown incorrectly
2. fix delegation modal height
3. fix delegation loader not shown

### DIFF
--- a/src/components/DelegateMultipleButton.jsx
+++ b/src/components/DelegateMultipleButton.jsx
@@ -43,7 +43,8 @@ const modalStyles = {
     transform: 'translate(-50%, -50%)',
     boxShadow: '0 0 40px #ccc',
     overflowY: 'auto',
-    height: '64%',
+    maxHeight: '64%',
+    minHeight: '350px',
   },
 };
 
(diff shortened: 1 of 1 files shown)


src/components/DelegateMultipleButton.jsx
```diff
@@ -187,6 +188,31 @@ const DelegateMultipleButton = props => {
...
+    if (sorted.length >= config.donationDelegateCountLimit) {
+      for (let i = 0; i < config.donationDelegateCountLimit; i += 1) {
+        sum += sorted[i]._amountRemaining.toNumber();
+      }
+    }
+    return sum;
+  };
+
+  const isLimitedDelegateCount = () => {
+    if (props.milestone && props.milestone.isCapped) {
```
aminlatifi, line 207: If milestone is not capped this alert maybe shown. Consider an uncapped milestone and a campaign which has donations of 12 different pledges

aminlatifi (changes requested):

src/components/DelegateMultipleButton.jsx
RamRamez: If milestone is not capped and campaign has donations of 12 different pledges, the message should be shown.

src/components/DelegateMultipleButton.jsx
```diff
@@ -187,6 +192,14 @@ const DelegateMultipleButton = props => {
     setLoadingDonations(false);
   };
 
+  const isLimitedDelegateCount = () => {
+    if (props.milestone && props.milestone.isCapped) {
+      return totalDonations > delegations.length && ifDelegationLimited;
```
aminlatifi, line 197: Imagine this case:
I have a campaign with 20 donations of 1 DAI value from different donators (so each one will be in different pledge), and a milestone which needs 16 DAI. Then `ifDelegationLimited` will be false and the message is not shown while it's incorrect

aminlatifi (changes requested):

src/components/DelegateMultipleButton.jsx
RamRamez: In this case ifDelegationLimited is true since newAmount considers config.donationDelegateCountLimit

aminlatifi:
Thanks @RamRamez

PR #1951, opened 2021-03-22
F 1486 Auto updating new donations for milestone, campaign and DAC views

RamRamez:
Issue #1486

Issue #1486, opened 2020-10-22
Need Donation Badge


papermache:
@aminlatifi has description.

markoprljic:
@aminlatifi Can you please describe this issue so that I know what needs to be done? Thanks!

aminlatifi:
@markoprljic Sorry for late response
It happens when donations are changed, user currently needs to refresh the page to see new donations (it should be fixed in future to the list of donations be updated too).
![image](https://user-images.githubusercontent.com/5684607/97077924-8e632d80-15f4-11eb-9155-ff7a786f7312.png)

aminlatifi:
> @markoprljic Sorry for late response
> It happens when donations are changed, user currently needs to refresh the page to see new donations (it should be fixed in future to the list of donations be updated too).
> ![image](https://user-images.githubusercontent.com/5684607/97077924-8e632d80-15f4-11eb-9155-ff7a786f7312.png)

I generated it by collectiong/dusbursing

markoprljic:
@aminlatifi So ... you mean that when I'm on the page and a new donation just happened it will be automatically updated? 
In such case we should show a "New" badge + highlight the incoming donation(s) in the list below? Is that correct?
Those new donations would then be considered as "seen" when the user refreshes the page?

aminlatifi:
> @aminlatifi So ... you mean that when I'm on the page and a new donation just happened it will be automatically updated? 
> In such case we should show a "New" badge + highlight the incoming donation(s) in the list below? Is that correct?
> Those new donations would then be considered as "seen" when the user refreshes the page?

Currently, when new donations are added the badge shows the number only and list is not updated.
Your story is great and results better user experience. If you design it we would implement

markoprljic:
@aminlatifi I created a mockup for this new donations notifications inside the new Leaderboard table (because I have it in Figma). You can apply it to the old table we currently have on the site, it's the same thing. 

Here's a preview 
<img width="1255" alt="Image 2020-12-09 at 8 54 06 am" src="https://user-images.githubusercontent.com/5704285/101600930-49585680-39fc-11eb-9e9c-a386bd8fc056.png">

aminlatifi:
@peymannaderi3 See whether leaderboard api supports feathers watching (monitor updates)

aminlatifi:
@RamRamez Please fix this for DACs, Campaigns and Milestones

aminlatifi:
[...]

PR #2055, opened 2021-04-20
F 1486 Auto updating new donations for milestone, campaign and DAC

RamRamez:
Issue #1486 and PR #1951


src/components/views/ViewCampaign.jsx
```diff
@@ -127,18 +147,25 @@ const ViewCampaign = ({ match }) => {
       .catch(() => {
         setNotFound(true);
       });
-  }, []);
```
aminlatifi, line 130: If donationsTotal is changed everything is fetched from scratch! the better solution is to just update donations/leader board list of items.

src/components/views/ViewCampaign.jsx
```diff
@@ -81,7 +83,7 @@ const ViewCampaign = ({ match }) => {
       loadFromScratch ? 0 : aggregateDonations.length,
       (_donations, _donationsTotal) => {
         setAggregateDonations(loadFromScratch ? _donations : aggregateDonations.concat(_donations));
```
aminlatifi, line 85: Array are mutable data types, so concatenating `aggregateDonations` doesn't update the reference. 
```suggestion
        setAggregateDonations(loadFromScratch ? _donations : [...aggregateDonations, ..._donations]);
```

aminlatifi (changes requested):

aminlatifi:
@RamRamez I just reviewed Campaign related view, for others do the same please.

src/components/views/ViewCampaign.jsx
RamRamez: The Balances component needs it to be loaded again

`<Balances entity={campaign} />`

src/components/views/ViewCampaign.jsx
RamRamez: Since `loadFromScratch` is `true`, it is not concatenating.

src/components/views/ViewCampaign.jsx
```diff
@@ -73,15 +75,18 @@ const ViewCampaign = ({ match }) => {
...
-  const loadMoreAggregateDonations = (loadFromScratch = false) => {
+  const loadMoreAggregateDonations = (
+    loadFromScratch = false,
+    donationsBatch = donationsPerBatch,
+  ) => {
     setLoadingDonations(true);
     AggregateDonationService.get(
       campaign.id,
-      donationsPerBatch,
+      donationsBatch || donationsPerBatch,
```
aminlatifi, line 85: This `|| donationPerBatch` is not necessary.

src/components/views/ViewCampaign.jsx
```diff
@@ -110,6 +116,23 @@ const ViewCampaign = ({ match }) => {
...
+        campaignId,
+        0,
+        0,
+        (_donations, _donationsTotal) => {
+          setDonationsTotal(_donationsTotal);
+        },
+        err => {
+          ErrorHandler(err, 'Some error on fetching campaign donations, please try later');
+        },
+        Donation.COMMITTED,
```
aminlatifi, line 131: When user do a donation, a new donation with Pending status is created. That one should be included in donation count  for better interactivity. I mean user should see the new badge immediately after donating. So the default condition of `{ $ne: Donation.FAILED }` in getDonations is enough and that parameter is not necessary.

src/services/CampaignService.js
```diff
@@ -174,24 +174,36 @@ class CampaignService {
...
+    const query = {
+      status: { $ne: Donation.FAILED },
+      $or: [{ intendedProjectTypeId: id }, { ownerTypeId: id }],
+      ownerTypeId: id,
+      isReturn: false,
+      $sort: { usdValue: -1, createdAt: -1 },
+      $limit,
+      $skip,
+    };
+    if (status) {
```
aminlatifi, line 199: As describe above this would not be necessary.

aminlatifi (commented):
@RamRamez Please do the same for ViewDAC as discussed for campaign

Issue #2063, opened 2021-04-25
Delegate button is always working in background


RamRamez:
Delegate button processes is always working in background without even clicking delegate button.

aminlatifi:
@MoeNick I will test it myself

PR #2064, opened 2021-04-25
H 2063 do delegation calculation just when user clicks the button

RamRamez:
issue #2063


RamRamez:
In this PR, Donate, Delegate and Multiple Delegate are refactored to do all calculations and requests, just on user click.

RamRamez:
In this PR, Donate, Delegate and Multiple Delegate components are refactored to do all calculations and requests, just on user click.

aminlatifi:
I opened a campaign and tried to donate, switched to a token with balance zero in my wallet (WBTC)

image

Issue #2098, opened 2021-05-02
Upgrade react dependencies


aminlatifi:
Upgrade create-react-app and feathers-reactive packages.

aminlatifi:
Maye solve #2055

PR #2102, opened 2021-05-02
F 2098 update packages

MohammadPCh:
#2098


MohammadPCh:
We must to update Netlify to use Yarn instead of npm.
I do this for GitHub test.

Issue #2108, opened 2021-05-03
Implement new designs: gas paid info and related alerts


MoeNick:
Related to : #2021 and ##2049

Alert:
https://www.figma.com/file/SuQbqtNjtRNrlYGAXijaMI/Giveth-TRACE?node-id=4967%3A0
https://lottiefiles.com/5527-alert-notification-character

Public gas information: 
https://www.figma.com/file/SuQbqtNjtRNrlYGAXijaMI/Giveth-TRACE?node-id=4961%3A0

aminlatifi:
> Related to : #2021 and ##2049
> 
> Alert:
> https://www.figma.com/file/SuQbqtNjtRNrlYGAXijaMI/Giveth-TRACE?node-id=4967%3A0
> https://lottiefiles.com/5527-alert-notification-character
> 
> Public gas information:
> https://www.figma.com/file/SuQbqtNjtRNrlYGAXijaMI/Giveth-TRACE?node-id=4961%3A0

So beautiful, thanks @markoprljic

PR #2166, opened 2021-05-16
F 2108 gas paid info with mobile design

RamRamez:
Issue #2108


aminlatifi:
image
There is difference between implementation and design
image

aminlatifi:
The campaign is slightly different
Implementation:
image

Design

image

aminlatifi:
Conversation view (right column) is shorter, can you make both columns even

image

Issue #2167, opened 2021-05-17
Edit Trace: pic size get back to their original size. 


MoeNick:
I resize a pic in body, when go back to edit mode, it get back to it\s original size. 

![image.png](https://images.zenhubusercontent.com/604f242c2819d5da6426ad03/92a55f7a-d363-4274-9f0b-462424b2fd89)

aminlatifi:
@MoeNick Please test all kind of traces

MoeNick:
Passed on payments - Brave
Not Passed on expences. Brave & Chrome

PR #2196, opened 2021-05-22
H 2167 pic size get back to their original size

MohammadPCh:
#2167 
We have a related error and i think ramin fixed it in #2195


src/components/views/EditPayment.jsx
```diff
@@ -56,15 +56,7 @@ function EditPayment(props) {
     recipientAddress: '',
     conversionRateTimestamp: undefined,
   });
-
```
aminlatifi, line 59: We didn't do it for other types of Trace! is it ok?

Issue #2052, opened 2021-04-20
Frontend implementation for search bar inside campaign. 


MoeNick:

MoeNick:
@aminlatifi it shouldn't get tested on UAT before closing?

MoeNick:
1- Won't search within characters on titles or body
2- Must show "no result " msg after search has no results
3- Must have search botton amd UX improvement, it's confusing to press magnifier icon while there is some other icons there.
4- It should be a "clear search" or reset button, especially after no result found and after each search done.

I think we'd better count on marko for this.

MoeNick:
Dear @markoprljic would u pls take a look, it needs a better design on:
1- placement of the bar, watermark and search button instead of magnifier
2- Remove search filters button.
3- Design "no result found"

Appreciate it.

markoprljic:
Can you please provide a screenshot of the Campaign page and where you want this search to be? 

What is a watermark? 
What search filters to remove? Where is that?

I'm lost.

aminlatifi:
> Can you please provide a screenshot of the Campaign page and where you want this search to be?
> 
> What is a watermark?
> What search filters to remove? Where is that?
> 
> I'm lost.

@markoprljic The search bar is implemented in develop.giveth.io campaign pages

markoprljic:
This one here?

![Image 2021-05-06 at 4 35 10 pm](https://user-images.githubusercontent.com/5704285/117316649-0e622e00-ae89-11eb-85ee-90bd5849f2af.jpg)

MoeNick:
> This one here?
> 
> ![Image 2021-05-06 at 4 35 10 pm](https://user-images.githubusercontent.com/5704285/117316649-0e622e00-ae89-11eb-85ee-90bd5849f2af.jpg)

Yes Man...

markoprljic:
This should be a live search, when user starts typing it automatically filters the results below.

In case no results matching the search then show this information below:

![Image 2021-05-06 at 5 12 59 pm](https://user-images.githubusercontent.com/5704285/117322510-5cc5fb80-ae8e-11eb-843f-fc64a63969d9.jpg)

Figma https://www.figma.com/file/SuQbqtNjtRNrlYGAXijaMI/Giveth-TRACE?node-id=4984%3A14
Lottie https://lottiefiles.com/5451-search-file

MoeNick:
> This should be a live search, when user starts typing it automatically filters the results below.
> 
[...]

PR #2198, opened 2021-05-23
F 2052 Frontend implementation for search bar inside campaign

RamRamez:
Issue #2052


aminlatifi:
Buttons and search input is different than design
image

image

aminlatifi:
The search file animation just should be played during search, and after the result is ready it should be a static image
image

RamRamez:
> The search file animation just should be played during search, and after the result is ready it should be a static image
> image

Done

aminlatifi:
Thanks @RamRamez. For the search input please implement `debounce` input. Currently by every button hit a new search is triggered.

Issue #2212, opened 2021-05-30
Gas paid info is shown for DAC


aminlatifi:
On develop

![image](https://user-images.githubusercontent.com/5684607/120095768-f2c30e00-c13c-11eb-84be-da7e5b9f8469.png)

MoeNick:
I cant test on develop so assume as done.

PR #2213, opened 2021-05-30
H 2212 hide gas paid value for DAC

RamRamez:
Issue #2212


src/components/Balances.jsx
```diff
@@ -93,11 +93,13 @@ const Balances = ({ entity }) => {
             </tbody>
           </table>
           <Row className="p-2 mb-4" justify="space-between" style={{ fontSize: '0.8rem' }}>
-            <Col>
```
aminlatifi, line 96: Please add a boolean attribute named `showGasUsed` and pass to component. It' not nice to be dependent on datatype in component

aminlatifi (changes requested):

aminlatifi (approved):

Issue #2223, opened 2021-05-31
After donating to a trace, (while the donation is still pending) the delegation of 3% shows up as a donation to the trace, but a separate donation


laurenluz:
Is this working as functioned? I donated to a Trace (on develop with brave browser) and the owner enabled "delegate 3% to community"... but both parts of the donations show up as donations to the Trace, just split up.

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/fff2db06-f4c2-4ff5-99b8-3fd3f3433dc9)

laurenluz:
UPDATE: after the donation transaction completed... only 1 part of the donation remained on the trace's page (as expect).

so my feedback is that perhaps we should have the pending transaction also not be displayed on the trace's page, but rather on the community's page, as expected

![image](https://user-images.githubusercontent.com/75490971/120239794-4e17fd80-c21c-11eb-9a70-230992f4c9cf.png)

mohammadranjbarz:
@MoeNick 

For testing this, please create a trace and mark the checkbox to donate 3% to Community, then after donating to trace you will see one donation to trace (minus 3%)

laurenluz:
it's working fine now on develop. Thanks Mohammad!

![image](https://user-images.githubusercontent.com/75490971/120846867-f46f4600-c52f-11eb-8631-e10e25885c57.png)

PR #2232, opened 2021-06-02
Make pending donation 3 percent to community correctly

mohammadranjbarz:
related to #2223

the pending donation for donate to Community (3 percent of donations) was creating incorrectly, so it was showing that donation in trace page but after mining transaction it was ok, but now we create pending transaction correctly

### DIFF
--- a/src/components/DonateButtonModal.jsx
+++ b/src/components/DonateButtonModal.jsx
@@ -346,24 +346,30 @@ const DonateButtonModal = props => {
     return clearUp;
   }, [canDonateToProject, getMaxAmount, updateAllowance, model]);
 
+  /**
+   *
+   * @param toAdmin
+   * {
+   *    adminId:number,
+   *    id:string <entityId>,
+   *    type: community | trace | campaign
+   *    campaignId ?: it's filled for traces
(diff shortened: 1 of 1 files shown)


src/components/DonateButtonModal.jsx
```diff
@@ -350,10 +350,20 @@ const DonateButtonModal = props => {
     adminId,
     _amount,
     donationOwnerAddress,
+    community,
```
aminlatifi, line 353: Let's remove `community` and replace `adminId` with a structure holding more complete data of receiver project

aminlatifi (changes requested):

src/components/DonateButtonModal.jsx
mohammadranjbarz: @aminlatifi 
It changed as you said

aminlatifi:
Thanks @mohammadranjbarz

src/components/DonateButtonModal.jsx
aminlatifi: @mohammadranjbarz Thanks

Issue #2172, opened 2021-05-17
initial donation amount in slider differ from one token to another


MoeNick:
Video sent to Ramin can't uplad here.

MoeNick:
Let's make all default amount to zero,

divine-comedian:
I'm not sure how to test this! I need more clarification!

mohammadranjbarz:
> I'm not sure how to test this! I need more clarification!

When you wanted to donate and dialogue was opening for `ETH` token the default amount was `1` but for other tokens it was zero, now we changed the default amount to  be zero for all tokens

PR #2238, opened 2021-06-05
Make initial amount zero for all donations tokens

mohammadranjbarz:
related to #2172


aminlatifi:
@mohammadranjbarz Default value is set in two other places in the same file. Please change them too.

mohammadranjbarz:
> @mohammadranjbarz Default value is set in two other places in the same file. Please change them too.

@aminlatifi I think current changes is enough, if you test it you'll see it works good now

Issue #2220, opened 2021-05-31
make 'edit trace' and 'delete trace' buttons more noticeable


divine-comedian:
as a user while viewing my trace I found it hard to find the buttons to edit and delete my trace, I found them eventually up in the top right corner but the text is too small and can be hard to see depending on the trace image. I suggest if we could increase the box-size, the text size/weight and add a background color to the button so it is more distinguishable on the page. 

![image.png](https://images.zenhubusercontent.com/5fff1c86d883b3d9a858ddbc/e4004471-4656-44d6-a39f-ab529d0b9a0a)

markoprljic:
For Trace Owner View, why not have it below the Trace title in header? They don't need the donate button, right?
Also, please note that the labels are now Edit and Delete.

![Image 2021-06-02 at 12 16 50 pm](https://user-images.githubusercontent.com/5704285/120463696-77f82e00-c39c-11eb-91e2-c7e1436df6e9.jpg)

MoeNick:
@RamRamez is it ok for you? Dont u want t say any comments about the background stuff that was mentioned in the daily?

RamRamez:
Thanks @markoprljic. Ghost button is a good idea.
I just want to add we have a gradient on background header to make text more clear on light background pictures.

aminlatifi:
> Thanks @markoprljic. Ghost button is a good idea.
> I just want to add we have a gradient on background header to make text more clear on light background pictures.

@RamRamez We have dark gradient on projects header already.

divine-comedian:
![image](https://user-images.githubusercontent.com/67759413/121239374-2a316900-c867-11eb-82b2-9960770d25dd.png)
much more noticeable! This looks great! I checked for all types of Traces and this looks to be fixed as proposed.

PR #2240, opened 2021-06-05
F 2220 move edit, cancel and delete buttons

RamRamez:
Issue #2220
move edit, cancel and delete buttons in trace, camp and community under title in header


aminlatifi (commented):
1. Cancel button is not shown in Campaign header, but it should
image

2. Edit button is shown for a Canceled Campaign which shouldn't
image

RamRamez:
> 1. Cancel button is not shown in Campaign header, but it should
>    image
> 2. Edit button is shown for a Canceled Campaign which shouldn't
>    image

Done

Issue #2254, opened 2021-06-08
Infinite unlock tooltip not showing completely


RamRamez:
![image.png](https://images.zenhubusercontent.com/6040fd6c6873c665b22c45b3/3c71e3c2-9ade-4439-bfc5-9479065078b4)

MoeNick:
@RamRamez do we still have this tooltip in our new design?

RamRamez:
> @RamRamez do we still have this tooltip in our new design?

No we don't have

Issue #2252, opened 2021-06-08
In the delegation screen and the number keeps changing


GriffGreen:
![Screen Shot 2021-06-07 at 9 10 23 PM](https://user-images.githubusercontent.com/14945613/121117104-d6763f80-c7d4-11eb-9ca1-cedf7a11f514.png)

I was donating 2351... and then i click donate and before i can click accept on the Trace and the number changes on me!!

Issue #2236, opened 2021-06-02
I delegated 3100 but only 1736.508 DAI showed up?!


GriffGreen:
https://beta.giveth.io/milestone/dan-trusted-seeds-gardener-core-team

<img width="951" alt="Screen Shot 2021-06-02 at 9 47 15 AM" src="https://user-images.githubusercontent.com/14945613/120511366-956ecd00-c387-11eb-8668-cde4b5287441.png">

GriffGreen:
oh weird... ok so his payment isnt stuck, but i delegated 3100, but only 1736.508 DAI showed up!

aminlatifi:
> oh weird... ok so his payment isnt stuck, but i delegated 3100, but only 1736.508 DAI showed up!

@GriffGreen Are you sure the value is not decreased due to scrolling?!! Like the video

https://user-images.githubusercontent.com/5684607/120616440-7a2abd00-c46e-11eb-8680-57174644d848.mov

mohammadranjbarz:
@GriffGreen  @aminlatifi 
I got a dump from the beta database then I run a simulation script to simulate transaction until the last block number before Griff's transaction, then I changed web3PRovider to login instead of Griff then I delegated `3100` from `Common stack society` to this milestone, and I saw that donations amount are ok and sum of them would be 3100, So I conclude that it was user error, I mean after scrolling the amount has changed and Griff didn't understand and signed the transaction with invalid amount

141.274 +450+450+450 +450+150+900+108.726 = 3100

![Screen Shot 1400-03-16 at 14.37.38.png](https://images.zenhubusercontent.com/5fa7ed7fb0994323d1a66034/735ffc44-64d7-4f4a-84e9-3bd207716ddc)

@MoeNick 
I think we can close this issue

aminlatifi:
> @GriffGreen @aminlatifi
> I got a dump from the beta database then I run a simulation script to simulate transaction until the last block number before Griff's transaction, then I changed web3PRovider to login instead of Griff then I delegated `3100` from `Common stack society` to this milestone, and I saw that donations amount are ok and sum of them would be 3100, So I conclude that it was user error, I mean after scrolling the amount has changed and Griff didn't understand and signed the transaction with invalid amount
> 
> 141.274 +450+450+450 +450+150+900+108.726 = 3100
> 
[...]

Issue #2181, opened 2021-05-17
The size of the pop up for the delegation isnt large enough to see the button to click when you are done


GriffGreen:
<img width="1136" alt="Screen Shot 2021-05-17 at 3 41 53 PM" src="https://user-images.githubusercontent.com/14945613/118565140-7c7ae080-b726-11eb-8603-78ddf6bc162e.png">

PR #2257, opened 2021-06-08
F 2236 Change donate and delegate modals to fit screen size

RamRamez:
Issue #2236 #2254 #2181 #2252
1. Change donate and delegate, delegate multiple modals to fit screen size
2. Change all donate modal components (select, input etc.) to ant
3. Fix Infinite unlock tooltip not showing completely

### DIFF
--- a/src/components/DelegateButton.jsx
+++ b/src/components/DelegateButton.jsx
@@ -13,7 +13,7 @@ import DelegateButtonModal from './DelegateButtonModal';
 
 const modalStyles = {
   minWidth: '60%',
-  maxWidth: '90%',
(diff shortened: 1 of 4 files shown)


aminlatifi:
Thanks @RamRamez, it is supper cool.
Just a point, please limit its max width. It is too wide in big screens.
Also I think slider can have some padding/margin from both side and will improve UI.
image

RamRamez:
> Thanks @RamRamez, it is supper cool.
> Just a point, please limit its max width. It is too wide in big screens.
> Also I think slider can have some padding/margin from both side and will improve UI.
> image

Done

Issue #2256, opened 2021-06-08
'your delegations' menu options are missing and the text is mispelled.


divine-comedian:
![image.png](https://images.zenhubusercontent.com/5fff1c86d883b3d9a858ddbc/1f29d7a1-6cb0-4a5b-85b7-6e6bb1afa7a0)

When I go to the my delegations page when in the DAC owner account I have the option to delegate funds that have been delegated to me. I have plenty of options of where to delegate my trace but there are no options in the campaigns menu for me to delegate... Also the text over the hover tooltip should read:

"Choose a Campaign or a Trace to delegate your funds to."

MoeNick:
Tested. Good job guys.

PR #2262, opened 2021-06-09
Remove filter of owner address in getting campaigns in delegation page

mohammadranjbarz:
related to #2256

Screen Shot 1400-03-19 at 16 33 22
Screen Shot 1400-03-19 at 16 33 39

### DIFF
--- a/src/components/DelegateButtonModal.jsx
+++ b/src/components/DelegateButtonModal.jsx
@@ -288,7 +288,6 @@ class DelegateButtonModal extends Component {
     } else {
       campaignValue.push(...objectsToDelegateToCampaign);
     }
-
     return (
       <React.Fragment>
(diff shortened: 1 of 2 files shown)


aminlatifi:
@mohammadranjbarz  It addresses the #2256, nice :)

But the queries only fetch 100 campaigns and traces as well. So if someday we have more, they will be omitted. What do you think?

mohammadranjbarz:
> @mohammadranjbarz It addresses the #2256, nice :)
> 
> But the queries only fetch 100 campaigns and traces as well. So if someday we have more, they will be omitted. What do you think?

@aminlatifi 
In my opinion:  if campaigns and traces grows a lot  fetching all of them in this page is not a good idea, maybe took a lot of time, so in that case when user type to search frontend should request to backend to get traces or campaigns(and we don't get them in the page loading)

PR #2280, opened 2021-06-14
Text fixes

divine-comedian:
Various text changes, grammar fixes and capitalizations.

### DIFF
--- a/src/models/Donation.js
+++ b/src/models/Donation.js
@@ -14,7 +14,7 @@ import Campaign from './Campaign';
  */
 class Donation extends Model {
   static get PENDING() {
-    return 'Pending';
+    return 'pending';
   }
 
   static get TO_APPROVE() {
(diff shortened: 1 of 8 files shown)


src/models/Donation.js
```diff
@@ -18,7 +18,7 @@ class Donation extends Model {
   }
 
   static get TO_APPROVE() {
-    return 'ToApprove';
+    return 'To Approve';
```
aminlatifi, line 21: No, you should not change it. It's mapped to status in backend.

aminlatifi (changes requested):

src/models/Donation.js
divine-comedian: fixed in commit below :point_down:

src/models/Donation.js
```diff
@@ -22,32 +22,32 @@ class Donation extends Model {
   }
 
   static get WAITING() {
-    return 'Waiting';
```
aminlatifi, line 25: These ones also should be changed.

aminlatifi (changes requested):

aminlatifi:
Thanks @divine-comedian

Issue #1681, opened 2021-01-17
Use UserProvider in all components instead of passing userInfo as a property


aminlatifi:
We have created a react provider for user info, but had passed it as property in many places. Those passed props should be removed and UserProvider should replace it.

RamRamez:
In PR #2296 the below issues are done:
1. Remove unnecessary props: currentUser, displayForeignNetRequiredWarning, balance, isForeignNetwork, match and history props
2. Fix Community Button wrong url type

aminlatifi:
Tested, it's ok

PR #2296, opened 2021-06-27
H 1681 Use UserProvider in all components

RamRamez:
Issue #1681
1. Remove unnecessary props: currentUser, displayForeignNetRequiredWarning, balance, isForeignNetwork, match and history props
2. Fix Community Button wrong url type

### DIFF
--- a/src/containers/Routes.jsx
+++ b/src/containers/Routes.jsx
@@ -59,19 +59,8 @@ const Routes = () => {
       {/* Routes are defined here. Persistent data is set as props on components
                                 NOTE order matters, wrong order breaks routes!
                             */}
-      <Route
-        exact
-        path="/communities/new"
-        render={props => (
-          <EditCommunity
-            isNew
-            balance={balance}
-            isForeignNetwork={isForeignNetwork}
-            displayForeignNetRequiredWarning={displayForeignNetRequiredWarning}
-            {...props}
-          />
-        )}
-      />
+      <Route exact path="/communities/new" render={props => <EditCommunity isNew {...props} />} />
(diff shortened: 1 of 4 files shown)


src/containers/Routes.jsx
```diff
@@ -92,58 +81,19 @@ const Routes = () => {
...
-          <EditCampaign
-            isNew
-            balance={balance}
-            isForeignNetwork={isForeignNetwork}
-            displayForeignNetRequiredWarning={displayForeignNetRequiredWarning}
-            {...props}
-          />
-        )}
-      />
+      <Route exact path="/campaigns/new" render={props => <EditCampaign isNew {...props} />} />
```
aminlatifi, line 84: If props are not needed they should be removed from these components too

src/containers/Routes.jsx
```diff
@@ -243,55 +193,15 @@ const Routes = () => {
...
         path="/my-communities"
-        render={props => (
-          <MyCommunities
-            key={currentUser ? currentUser.id : 0}
-            currentUser={currentUser}
-            balance={balance}
-            {...props}
-          />
-        )}
+        render={() => <MyCommunities currentUser={currentUser} balance={balance} />}
```
aminlatifi, line 201: Let's remove currentUser and balance from these components too, it means making them Hook Base components

aminlatifi (changes requested):

src/containers/Routes.jsx
RamRamez: EditCampaign needs match props

Issue #2320, opened 2021-06-30
Co-owners cannot edit traces


freshelle:
![image](https://user-images.githubusercontent.com/86718721/123946930-53f72f00-d9d2-11eb-825a-611848917f02.png)

I clicked the "edit" button and the screen flashes, and it won't let me edit traces.

freshelle:
Hi Amin. I see that the issue is still open. I tried to edit a trace again today and I experienced the same issue. I'll keep posted for updates on this. Thank you so much!

aminlatifi:
> Hi Amin. I see that the issue is still open. I tried to edit a trace again today and I experienced the same issue. I'll keep posted for updates on this. Thank you so much!

Hi @freshelle
We are working on it, it will be fixed on develop and then will be released to production env. We hope it happen soon this week

MoeNick:
@RamRamez @aminlatifi 

**Tested, as co-owner I can:**
1- Edit, Accept, reject. from trace page and from dashboard. 
2- All Traces type: Expense, Bounty and Payment. 

Problems:
1. Wrong update msg on edit payment and expense: It says Bounty is get updated! 
2. If not signed in, needs double click on reject, edit and re-propose botton. 
3.  edit trace small badge over the trace card not showed on campaign page for owner and co-owner.

RamRamez:
> @RamRamez @aminlatifi
> 
> **Tested on Brave, as co-owner I can:**
> 1- Edit, Accept, reject. from trace page and from dashboard.
> 2- All Traces type: Expense, Bounty and Payment.
> 
> Problems:
> 
> 1. Wrong update msg on edit payment and expense: It said Bounty is get updated!
> 2. If not signed in, needs double click on reject, edit and re-propose botton.
> 3. edit trace small badge over the trace card not showed on campaign page for owner and co-owner.

No. 3: Only proposed traces can be edited by campaign owner or campaign coowner and we do not show proposed traces on campaign page

RamRamez:
> @RamRamez @aminlatifi
> 
> **Tested on Brave, as co-owner I can:**
> 1- Edit, Accept, reject. from trace page and from dashboard.
> 2- All Traces type: Expense, Bounty and Payment.
> 
> Problems:
> 
> 1. Wrong update msg on edit payment and expense: It said Bounty is get updated!
> 2. If not signed in, needs double click on reject, edit and re-propose botton.
> 3. edit trace small badge over the trace card not showed on campaign page for owner and co-owner.

[...]

PR #2321, opened 2021-06-30
Update edit allowed condition in traces

aminlatifi:
Refs #2320


aminlatifi:
@RamRamez I didn't test the patch. Probably it will have some issue in state management. Please test and fix any issue if exists.

RamRamez:
1. Fix state management in setting owner
2. Show notification when user not allowed to edit

aminlatifi:
LGTM

Issue #2319, opened 2021-06-30
Use Ant components in ChangeOwnershipButton component 


aminlatifi:

RamRamez:
Done in PR #2322 

1. Remove react-rangeslider component
2. Use ant components in change ownership
3. Add ability to remove Campaign co-owner

MoeNick:
Any Ideas for default value of changing co-owner modal? 

Eider a campaign has a co-owner or not. so putting the owner address as a default value in dropdown list is a bad idea I think. 
What do you think? 
@RamRamez

RamRamez:
> Any Ideas for default value of changing co-owner modal?
> 
> Eider a campaign has a co-owner or not. so putting the owner address as a default value in dropdown list is a bad idea I think.
> What do you think?
> @RamRamez

There was a problem before that the campaign owner could make himself co-owner, so in this PR I removed Campaign owner from co-owner selection list and this is not also a default selection. The default selection is the previous campaign co-owner so in your case, probably the campaign owner had been set as campaign co-owner in DB before.

MoeNick:
So is it okay that I keep this issue here to see your recent changes and after move it to done? 
@RamRamez

RamRamez:
> So is it okay that I keep this issue here to see your recent changes and after move it to done?
> @RamRamez

It seems when creating campaign, backend is setting the default co-owner address equal to the campaign owner.
@mohammadranjbarz Could you please check it?

mohammadranjbarz:
> > So is it okay that I keep this issue here to see your recent changes and after move it to done?
> > @RamRamez
> 
> It seems when creating campaign, backend is setting the default co-owner address equal to the campaign owner.
> @mohammadranjbarz Could you please check it?

Yes you are right
by default the `coownerAddress` is the userAddress of who create the campaign

MoeNick:
I created another issue and make it done.

RamRamez:
> > > So is it okay that I keep this issue here to see your recent changes and after move it to done?
> > > @RamRamez
> > 
> > 
> > It seems when creating campaign, backend is setting the default co-owner address equal to the campaign owner.
> > @mohammadranjbarz Could you please check it?
> 
> Yes you are right
> by default the `coownerAddress` is the userAddress of who create the campaign

@MoeNick has created an issue for it #2339

PR #2322, opened 2021-06-30
F 2319 use ant components in change ownership button

RamRamez:
Related to #2319

1. Remove react-rangeslider component
2. Use ant components in change ownership
3. Add ability to remove Campaign co-owner

### DIFF
--- a/src/components/ChangeOwnershipButton.jsx
+++ b/src/components/ChangeOwnershipButton.jsx
@@ -1,42 +1,23 @@
-import React, { Fragment, useContext, useState, useRef } from 'react';
-import Modal from 'react-modal';
+import React, { Fragment, useContext, useState, useEffect } from 'react';
 import BigNumber from 'bignumber.js';
-import { Form } from 'formsy-react-components';
 import PropTypes from 'prop-types';
-import 'react-rangeslider/lib/index.css';
+import { Button, Modal, Select, Form } from 'antd';
 
 import Campaign from 'models/Campaign';
-import Trace from 'models/Trace';
-import { Button } from 'antd';
-import { ZERO_ADDRESS } from '../lib/helpers';
 import { isLoggedIn, checkBalance } from '../lib/middleware';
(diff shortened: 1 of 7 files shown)


src/components/ChangeOwnershipButton.jsx
```diff
@@ -1,42 +1,23 @@
...
-    bottom: 'auto',
-    marginRight: '-20%',
-    transform: 'translate(-50%, -50%)',
-    boxShadow: '0 0 40px #ccc',
-    overflowY: 'scroll',
-  },
-};
+
+const noCoownerValue = '------ No co-owner ------';
+const noCoownerObject = { value: noCoownerValue, title: noCoownerValue };
```
aminlatifi, line 20: value should be zero address probably

src/components/ChangeOwnershipButton.jsx
```diff
@@ -87,23 +69,39 @@ const ChangeOwnershipButton = props => {
...
+      coownerAddress === noCoownerObject.value ? ZERO_ADDRESS : coownerAddress,
       afterCreate,
       afterMined,
     );
   };
 
+  // Set initial value for select component
+  useEffect(() => {
+    if (campaignManagers.length > 0) {
+      if (campaign.coownerAddress > ZERO_ADDRESS) {
```
aminlatifi, line 81: They are string! that's comparison is not solid! use 
`!==`

src/components/ChangeOwnershipButton.jsx
```diff
@@ -87,23 +69,39 @@ const ChangeOwnershipButton = props => {
       campaign,
       currentUser.address,
       campaign.ownerAddress,
-      campaign.coownerAddress,
+      coownerAddress === noCoownerObject.value ? ZERO_ADDRESS : coownerAddress,
```
aminlatifi, line 72: what does it mean?

aminlatifi (changes requested):

Issue #1538, opened 2020-11-23
Add onboardjs or alternative to Giveth 1 


aminlatifi:
Replace metamsk only wallet with onboardjs

PR #2337, opened 2021-07-06
F 1538 Implement onboard.js

RamRamez:
Issue #1538


src/services/DonationBlockchainService.jsx
```diff
@@ -490,54 +471,50 @@ class DonationBlockchainService {
    * @param {string}   address   Address of the user who calls commit
    * @param {function} onCreated Callback function after the transaction has been broadcasted to chain and stored in feathers
    * @param {function} onSuccess Callback function after the transaction has been mined
-   * @param {function} onError   Callback function after error happened
+   * @param liquidPledging       liquidPledging
    */
-  static commit(donation, address, onCreated = () => {}, onSuccess = () => {}, onError = () => {}) {
+  static commit(donation, address, onCreated = () => {}, onSuccess = () => {}, liquidPledging) {
```
aminlatifi, line 476: Please pass web3 like other places in project.

src/contextProviders/Web3Provider.jsx
```diff
@@ -293,12 +107,13 @@ class Web3Provider extends Component {
...
       isHomeNetwork,
       isForeignNetwork,
-      isEnabled,
-      setupTimeout,
       showForeignNetRequiredWarning,
       showHomeNetRequiredWarning,
+      web3,
     } = this.state;
 
+    const isEnabled = !!account && !!balance;
```
aminlatifi, line 115: Is `&& !!balance` needed here?

aminlatifi (changes requested):
Thanks @RamRamez big improvement

src/contextProviders/Web3Provider.jsx
RamRamez: We have four subscriptions: web3, network, address and balance. First web3 become ready and last balance will be ready. On each of these four, App will be rerendered by Web3 Context Provider. I think `!!balance` is enough but to make sure we can include all four subscriptions in that conditions.

Issue #1590, opened 2020-12-09
My Account / Manage section - Redesign


markoprljic:
My Account - Manage section for Donations, Milestones, Delegations (Campaigns as well?)

Started the mockups to structure the menu and main sections of the Manage pages.

![Image 2020-12-09 at 8.57.36 am.png](https://images.zenhubusercontent.com/5cc090818b624e26c3734d75/82dc0972-707b-4f9f-be59-d88e7b2adef4)

MoeNick:
@RamRamez @MohammadPCh please go forward and ask marko if its not clear to you.

divine-comedian:
testing on `develop.giveth.io` 

new dashboard looks great and the tabs work very fluid
![image](https://user-images.githubusercontent.com/67759413/126492887-5734e996-9743-4df4-bedb-9e793d04e0c5.png)

looks like this is resolved!

RamRamez:
We still need two user actions to show this new dashboard: First user should hover mouse on his profile name, Then he should click on one of the menus.
@markoprljic It would be great if we could show this dashboard with just one click (e.g. just clicking on the name and remove that dropdown menu).

divine-comedian:
I was doing more testing and just uncovered a pretty big issue - notifications are popping up behind the dashboard bar, making them almost impossible to see - 

![image](https://user-images.githubusercontent.com/67759413/126518306-14dbc938-5319-44df-97b5-29a62d599052.png)

RamRamez:
> I was doing more testing and just uncovered a pretty big issue - notifications are popping up behind the dashboard bar, making them almost impossible to see -
> 
> ![image](https://user-images.githubusercontent.com/67759413/126518306-14dbc938-5319-44df-97b5-29a62d599052.png)

Fixed in PR #2406

MoeNick:
Thanks @divine-comedian  and @RamRamez 

notifications problem seems have been solved.

PR #2406, opened 2021-07-21
H 1590 fix toast showing under menu

RamRamez:
Issue #1590


MohammadPCh:
Dear @RamRamez ,
I think it'b better to set z-index to some-think like 1030.
We should have policy for our z-indexes.
It help us to handle layers in the future :)
Thanks.

MohammadPCh (changes requested):
Maybe 1030 is better :)

RamRamez:
> Maybe 1030 is better :)

Yeah that's better, Done.

Issue #2364, opened 2021-07-11
Add Source Map for Sentry


MoeNick:

MoeNick:
related to #2356

This is working but we need add sourcemap to get variables and codes real names in sentry reports not uglified names, but I was not familiar with react webpack config ( I don't know without rejecting, how we can modify webpack config) maybe @MohammadPCh or @RamRamez can help,

https://docs.sentry.io/platforms/javascript/guides/react/sourcemaps/

@aminlatifi and we should add REACT_APP_SENTRY_DSN in netlify when building

PR #2440, opened 2021-08-01
F 2364 add react sourcemap to sentry

MohammadPCh:
Please add these variables to the environment and user **yarn release** instead of yarn build. 
REACT_APP_SENTRY_DSN
SENTRY_AUTH_TOKEN
SENTRY_ORG
SENTRY_PROJECT
#2364


src/index.js
```diff
@@ -1,19 +1,19 @@
 import React from 'react';
 import localForage from 'localforage';
 import { hydrate, render } from 'react-dom';
-import * as Sentry from '@sentry/react';
-import { Integrations } from '@sentry/tracing';
+import * as Sentry from '@sentry/browser';
```
aminlatifi, line 4: Do we need to use `@sentry/browser` and add dependency? I mean can we use the same '@sentry/react' and '@sentry/tracing'

src/index.js
```diff
@@ -1,19 +1,19 @@
...
 import './lib/SegmentAnalytics';
-import config from './configuration';
 
-Sentry.init({
-  dsn: config.sentryDsn,
-  integrations: [new Integrations.BrowserTracing()],
-  tracesSampleRate: 0.1,
-});
+// Setup setnry
+if (process.env.NODE_ENV === 'production' && process.env.REACT_APP_SENTRY_RELEASE) {
```
aminlatifi, line 11: Why it will only be used on `production`? I had set config on develop env

MohammadPCh:
Use sentry react
add test route
fix to work in development too.
@aminlatifi

Issue #2441, opened 2021-08-01
Refactor old trace edit component


RamRamez:
Refactor old trace edit component and fix some UI and functional bugs

PR #2442, opened 2021-08-01
F 2397 remove formsy

RamRamez:
Issue #2441 #2397

1. Remove formsy
2. Complete refactor EditTraceOld


src/components/EditTraceCommons.jsx
```diff
@@ -234,16 +235,15 @@ const TraceDatePicker = ({ onChange, value, disabled }) => {
   return (
     <Row gutter={16}>
       <Col className="gutter-row" span={10}>
-        <Form.Item label="Date" className="custom-form-item">
+        <Form.Item
+          label="Date"
+          className="custom-form-item"
+          rules={[{ required: true, message: 'Date is required' }]}
+        >
```
aminlatifi, line 242: image

I submit the trace without date and there is no error in form, and saving threw error in notification!

aminlatifi (changes requested):

src/components/EditTraceCommons.jsx
RamRamez: We had this problem with other trace types (bounty, etc) because ant doesn't have required mark for datePicker. I solved it by adding default picker value for date.

Issue #2182, opened 2021-05-17
Hard to see the comment when I delegate and add a message, it shows up on the Trace.. but its hidden


GriffGreen:
<img width="991" alt="Screen Shot 2021-05-17 at 4 12 30 PM" src="https://user-images.githubusercontent.com/14945613/118567490-bfd74e00-b72a-11eb-91c0-6ee0079c8da0.png">
<img width="598" alt="Screen Shot 2021-05-17 at 4 11 47 PM" src="https://user-images.githubusercontent.com/14945613/118567504-c4036b80-b72a-11eb-9714-4e37f878e751.png">
<img width="926" alt="Screen Shot 2021-05-17 at 4 15 54 PM" src="https://user-images.githubusercontent.com/14945613/118567750-383e0f00-b72b-11eb-913d-efca509e7ccf.png">

mosaeedi:
@MoeNick

PR #2446, opened 2021-08-02
H 2182 donation comment not shown in trace table

RamRamez:
Issue #2182

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "doc": "jsdoc src -r -d docs",
     "preci": "rm -fr node_modules",
     "ci": "yarn install --frozen-lockfile",
-    "release": "(export REACT_APP_SENTRY_RELEASE=$(git rev-parse --short HEAD); react-scripts build && node scripts/sentry.js)"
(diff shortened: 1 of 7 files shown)


aminlatifi:
image

I guess it would better to align other columns content vertically top. Also for long comments like these one `load more` button seems to be necessary.

RamRamez:
> image
> 
> I guess it would better to align other columns content vertically top. Also for long comments like these one `load more` button seems to be necessary.

Done

aminlatifi:
It should be blue basically
image

aminlatifi:
Donation table is wider and it makes it uneven
image
image

RamRamez:
> Donation table is wider and it makes it uneven
> image
> image

I intentionally made wider to remove horizontal scroll bar

aminlatifi:
Style change has caused some issues in donation table view. 
One is this:
telegram-cloud-photo-size-4-5807760584626779982-y
To have a sense, our current implementation is this:
telegram-cloud-photo-size-4-5807760584626779995-y

Another issue is when the donation table is not empty, it's uneven with rest of page as I mentioned above.

I suggest changing table like this:
image

1. I removed empty not used first column (an empty column has been remained from past and I don't know why!)
2. I removed address column. User address and etherscan link can be obtained by just one click on user name link (4th column)

Issue #2152, opened 2021-05-11
Show usdValue of donation amount in donate/delegate menu


mohammadranjbarz:
I think it's cool to show the usd value of the amount, it makes more sense for user to know how much he/she is donating

MoeNick:
@markoprljic what do you think?
![image.png](https://images.zenhubusercontent.com/604f242c2819d5da6426ad03/c240da8f-b1b3-47e3-9dd0-dd14a652eda9)

markoprljic:
Yes

mohammadranjbarz:
@MoeNick 
I think maybe it's better to assign this issue to @RamRamez  or @MohammadPCh

RamRamez:
These are done in PR #2447 (For UAT test):
1. Show usd Value of amount in donate, delegate and delegate multiple modal
2. Remove unused scss files
3. Make donate, delegate and delegate multiple modal responsive

PR #2447, opened 2021-08-03
F 2152 show usd value of donation amount

RamRamez:
Issue #2152 

These are done (For UAT test):
1. Show usd Value of amount in donate, delegate and delegate multiple modal
2. Remove unused scss files
3. Make donate, delegate and delegate multiple modal responsive


src/components/views/myDelegations/DelegateButtonModal.jsx
```diff
@@ -61,6 +63,25 @@ class DelegateButtonModal extends Component {
     this.selectedObject = this.selectedObject.bind(this);
   }
 
+  componentDidMount() {
```
aminlatifi, line 66: Please move the rate thing to only one hook, like `useCampaign`. The code is used three times

aminlatifi (changes requested):

src/components/views/myDelegations/DelegateButtonModal.jsx
aminlatifi: Nice job @RamRamez

Issue #2485, opened 2021-08-18
Update giveth packages


aminlatifi:
Giveth packages like giveth-bridge, giveth-common-contracts, etc. are not configured well and have old or unnecessary dependencies. This blocks us from upgrading our projects.

MoeNick:
@aminlatifi Would i pls set estimates?

PR #2489, opened 2021-08-18
F change bridge dependency

aminlatifi:
#2485


RamRamez (approved):
Thanks @aminlatifi 
I tested the bridge related tasks with node v14.17.5 and I couldn't find any bugs.

Issue #2469, opened 2021-08-13
No Campaigns show to Delegate funds to. 


divine-comedian:
Testing on develop.giveth.io

I'm a Community owner and I have a pile of funds that have been donated to me. I'm trying to delegate them to a Campaign from the My Delegations menu. I arrive on the screen where I can decide to donate to a campaign or trace and there are no options for campaigns for me to delegate to. 

![image.png](https://images.zenhubusercontent.com/5fff1c86d883b3d9a858ddbc/b2689733-85c6-4577-9e09-cf8a5c996900)

However if I go directly to a campaign I'm able to delegate funds on the Campaign page as a Community Owner. 

![image.png](https://images.zenhubusercontent.com/5fff1c86d883b3d9a858ddbc/6c1fc92a-7e5b-4b5e-af7d-5985802e9093)

I should be able to see a list of campaigns and delegate directly to them from the My Delegations menu.

MoeNick:
@RamRamez jan
Pls take a look. I forgot to report this one.

RamRamez:
> Testing on develop.giveth.io
> 
> I'm a Community owner and I have a pile of funds that have been donated to me. I'm trying to delegate them to a Campaign from the My Delegations menu. I arrive on the screen where I can decide to donate to a campaign or trace and there are no options for campaigns for me to delegate to.
> 
> ![image.png](https://camo.githubusercontent.com/c4f8138088a5de4d2c13a063293ceb0bed55df08d540246f67ac8e44fc99911b/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3566666631633836643838336233643961383538646462632f62323638393733332d383563362d343537372d396530392d636638613563393936393030)
> 
> However if I go directly to a campaign I'm able to delegate funds on the Campaign page as a Community Owner.
> 
> ![image.png](https://camo.githubusercontent.com/afec8a1f59851c87ac3142c0a3a6c29441670c492664bcd912d918021147b80c/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3566666631633836643838336233643961383538646462632f36633166633932612d376535622d346235652d616637642d353938353830326539303933)
> 
> I should be able to see a list of campaigns and delegate directly to them from the My Delegations menu.

I couldn't reproduce the bug
@divine-comedian Could you please check again to see whether you can reproduce the bug or not?

divine-comedian:
[...]

PR #2521, opened 2021-09-01
F 2469 choose anything in my delegations

aminlatifi:
Issue #2469

### DIFF
--- a/src/components/layout/MainMenu/RightMenu.jsx
+++ b/src/components/layout/MainMenu/RightMenu.jsx
@@ -84,7 +84,7 @@ const RightMenu = () => {
   };
 
   const getUserDelegations = () => {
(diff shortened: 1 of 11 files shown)


aminlatifi:
@RamRamez Let's the index be ready and then test it

aminlatifi:
@RamRamez Good job
just few tweaks,
1. When the campaign field is empty it's better to hide the checkbox and trace inpu
image
2. When the checkbox is enabled, the trace field should be required
image

RamRamez:
> @RamRamez Good job
> just few tweaks,
> 
> 1. When the campaign field is empty it's better to hide the checkbox and trace input
>    image
> 2. When the checkbox is enabled, the trace field should be required
>    image

Done

Issue #2470, opened 2021-08-13
Donation history is not shown correctly


aminlatifi:
This donation history representation is wrong and doesn't match with real history in DB

![image](https://user-images.githubusercontent.com/5684607/129415669-825b107f-6402-4f6b-80a0-1aef6168f320.png)

Note: donation history is obtained by traversing parentDonations field of donation enitity

MoeNick:
Who is able to take the bug? Is it front-end? @aminlatifi

aminlatifi:
> Who is able to take the bug? Is it front-end? @aminlatifi 

It's a front-end task but needs good understanding of donation entity fields. I guess @mohammadranjbarz can fix it faster since he has background knowledge

mohammadranjbarz:
> This donation history representation is wrong and doesn't match with real history in DB
> 
> ![image](https://user-images.githubusercontent.com/5684607/129415669-825b107f-6402-4f6b-80a0-1aef6168f320.png)
> 
> Note: donation history is obtained by traversing parentDonations field of donation entity

https://trace.giveth.io/trace/cherik-reward-may-1-aug-5

PR #2538, opened 2021-09-13
Fix donation history in trace page

mohammadranjbarz:
related to #2470

**Before**

donation history

**After**

Screen Shot 1400-06-22 at 10 24 33

### DIFF
--- a/src/components/DonationHistory.jsx
+++ b/src/components/DonationHistory.jsx
@@ -61,7 +61,7 @@ class DonationHistory extends Component {
       historyItem.parentIds,
       historyItem,
     );
-    if (parents.length > 0) {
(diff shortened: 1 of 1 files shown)


src/components/DonationHistory.jsx
```diff
@@ -61,7 +61,7 @@ class DonationHistory extends Component {
       historyItem.parentIds,
       historyItem,
     );
-    if (parents.length > 0) {
+    if (!historyItem.homeTxHash && parents.length > 0) {
```
mohammadranjbarz, line 64: If donation has `homeTxHash` we guess it is direct donation, so we should not check donation's parent

Issue #2541, opened 2021-09-22
Add switch network


RamRamez:
1. Add ability to automatically switch network in MetaMask when user is on wrong network
2. Unify wrong network warning modals
3. Remove switch network gif guide

PR #2542, opened 2021-09-22
H 2541 Add ability to automatically switch network

RamRamez:
Issue #2541 

1. Add ability to automatically switch network in MetaMask when user is on wrong network
2. Unify wrong network warning modals
3. Remove switch network gif guide


aminlatifi:
@MoeNick Can you accept the UI/UIX?

MoeNick:
Seems great to me. Do u have any objections that i missed?

aminlatifi:
> Seems great to me. Do u have any objections that i missed?

Not, it's cool

src/contextProviders/Web3Provider.jsx
```diff
@@ -77,34 +69,21 @@ const Web3Provider = props => {
...
-      showHomeNetRequiredWarning: true,
-      onHomeNetWarningClose: onClose,
-      homeNetWarningButtonLabel: buttonLabel,
-    });
+  const switchNetwork = () => {
+    if (isMetaMask) {
+      window.ethereum.request({
+        method: 'wallet_switchEthereumChain',
+        params: [
+          { chainId: isForeignNetwork ? config.homeNetworkChainId : config.foreignNetworkChainId },
```
aminlatifi, line 81: This implementation has issue when user wallet is connected to a third network (not home or foreign) and tries to donates! In this case the wallet will try to connect to foreign network instead of home network.

I suggest changing `ActionNetworkWarning` to accept an enum type parameter which indicates the aimed network

src/components/ActionNetworkWarning.jsx
```diff
@@ -1,30 +1,59 @@
...
       </div>
     )
   );
 }
 
 ActionNetworkWarning.propTypes = {
   incorrectNetwork: PropTypes.bool.isRequired,
-  networkName: PropTypes.string.isRequired,
+  networkName: PropTypes.string,
+  switchNetwork: PropTypes.func,
```
aminlatifi, line 47: Since `switchNetwork` is constant (the one obtained from web3 provider), we can remote this parameter

src/components/ActionNetworkWarning.jsx
```diff
@@ -1,30 +1,59 @@
...
+        {!isInline && switchNetworkBtn()}
       </div>
     )
   );
 }
 
 ActionNetworkWarning.propTypes = {
   incorrectNetwork: PropTypes.bool.isRequired,
-  networkName: PropTypes.string.isRequired,
+  networkName: PropTypes.string,
```
aminlatifi, line 46: We can remove this parameter as well if we decide to pass the required network in another parameter

aminlatifi (changes requested):

src/components/ActionNetworkWarning.jsx
RamRamez: I can't use web3provider context here because it causes a dependency cycle

image

src/components/ActionNetworkWarning.jsx
```diff
@@ -1,30 +1,59 @@
...
+import Web3 from 'web3';
+import config from '../configuration';
 
 /**
  * It is used to show in modals when user wants to take an action while connected to wrong network.
  */
 
 function ActionNetworkWarning(props) {
-  const { incorrectNetwork, networkName } = props;
+  const { incorrectNetwork, networkName, switchNetwork, web3, isInline } = props;
```
aminlatifi, line 12: Why we are passing such amount of parameters to this component? I mean all the network parameters can be summarized in one required chain id, the the component itself obtains all others (web3, network name, etc.) itself.

aminlatifi (changes requested):

PR #2549, opened 2021-09-27
F 2544 ability to withdraw any tokens bigger than withdraw limit

RamRamez:
Issue #2544

### DIFF
--- a/src/components/TraceActions.jsx
+++ b/src/components/TraceActions.jsx
@@ -24,13 +24,15 @@ function TraceActions({ trace }) {
     actions: { convertMultipleRates },
   } = useContext(ConversionRateContext);
   const {
-    state: { minimumPayoutUsdValue },
+    state: { minimumPayoutUsdValue, activeTokenWhitelist },
   } = useContext(WhiteListContext);
   const {
     state: { currentUser },
   } = useContext(UserContext);
   const [currentBalanceUsdValue, setCurrentBalanceUsdValue] = useState(0);
-  const [isAmountEnoughForWithdraw, setIsAmountEnoughForWithdraw] = useState(true);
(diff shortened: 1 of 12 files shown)


mohammadranjbarz:
@RamRamez 
You didnt handle withdraw when user want to withdraw from /my-traces

Screen Shot 1400-07-06 at 17 03 25

MoeNick:
Do you have any suggestion @mohammadranjbarz ?

RamRamez:
@mohammadranjbarz 
When there is only one token to withdraw, that token is selected by default and select component is disabled.
I can remove select component in this case and just a text instead

mohammadranjbarz:
> @mohammadranjbarz When there is only one token to withdraw, that token is selected by default and select component is disabled. I can remove select component in this case and put just a text instead

@MoeNick 
What do you think about it ?

RamRamez:
> @RamRamez You didnt handle withdraw when user want to withdraw from /my-traces
> 
> Screen Shot 1400-07-06 at 17 03 25

Yes you're right. I fixed it.

mohammadranjbarz:
> > @RamRamez You didnt handle withdraw when user want to withdraw from /my-traces
> > Screen Shot 1400-07-06 at 17 03 25
> 
> Yes you're right. I fixed it.

It's ok now
but I get an error when want to disburse money

Screen Shot 1400-07-07 at 09 42 04 (2)

RamRamez:
@mohammadranjbarz I simulated the same situation and it was ok for me. Maybe it was a MetaMask error or something. Can you test again?

mohammadranjbarz (approved):
Thanks, Looks good to me

mohammadranjbarz (approved):

mohammadranjbarz (approved):
@RamRamez 
Thanks, it seems good

